PR #1, opened 2018-01-16
Cleanup

dlsniper:
First of all, thank you for releasing this!

Second, here's a small PR to cleanup some things detected by GoLand.

There are a few more issues, also reported by https://github.com/golang/lint, but aside from the innocent documentation for exported fields, some of them would be breaking changes (depending on when you want to upgrade this, it might not be possible to address them now).

Hope it helps.


dlsniper:
I've also added a missing error handling in ` lambda.Start `

dlsniper:
@bmoffatt friendly ping :)

bmoffatt:
Thanks for the cleanup!

> some of them would be breaking changes

I'm assuming this would be in reference to `lambda/messages`? Yes, these need to remain stable indefinitely.

PR #9, opened 2018-01-16
Fix panic in handler.validateReturns()

### LINKED ISSUE #3: handler.validateReturns() panics on function with no return value
The validateReturns method panics with runtime: index out of range for a handler function with no return value. Tested with go v1.9.2

Adding this test case in `lambda/handler_test.go`:
<pre>
{
	name:     "no return value should not result in error",
	expected: nil,
	handler: func() {
	},
},
</pre>
results in
<pre>
panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range

goroutine 18 [running]:
testing.tRunner.func1(0xc420152b40)
	/usr/local/go/src/testing/testing.go:711 +0x2d2
panic(0x138f900, 0x15f9800)
	/usr/local/go/src/runtime/panic.go:491 +0x283
reflect.(*rtype).Out(0x1369c00, 0x0, 0x1393940, 0xc420016a88)
	/usr/local/go/src/reflect/type.go:1043 +0xc7
github.com/aws/aws-lambda-go/lambda.validateReturns(0x15d3040, 0x1369c00, 0x1369c00, 0x0)
	/Users/mattman/go/src/github.com/aws/aws-lambda-go/lambda/handler.go:64 +0x1c6
</pre>

Line 63 in handler.go needs to check for at least one return value before performing the type check.
<pre>
	} else if handler.NumOut() > 0 {
		if !handler.Out(0).Implements(errorType) {
</pre>

dlsniper:
This fixes a panic when invoking validateReturns() on a function without returns. Fixes #3

### DIFF
--- a/lambda/handler.go
+++ b/lambda/handler.go
@@ -58,7 +58,7 @@ func validateReturns(handler reflect.Type) error {
 		if !handler.Out(1).Implements(errorType) {
 			return fmt.Errorf("handler returns two values, but the second does not implement error")
 		}
-	} else {
+	} else if handler.NumOut() == 1 {
 		if !handler.Out(0).Implements(errorType) {
 			return fmt.Errorf("handler returns a single value, but it does not implement error")
 		}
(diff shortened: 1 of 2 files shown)


bmoffatt:
Thanks for the fix!

PR #11, opened 2018-01-16
Adds custom authorizer context to APIGW Request

deluan:
This PR enables parsing of the [Custom Authorizer Context](https://docs.aws.amazon.com/apigateway/latest/developerguide/use-custom-authorizer.html#api-gateway-custom-authorizer-output) as a map. 

The previous implementation was based on the example found in the link above, and did not allow for custom keys.

### DIFF
--- a/events/apigw.go
+++ b/events/apigw.go
@@ -33,6 +33,7 @@ type APIGatewayProxyRequestContext struct {
 	RequestID    string                    `json:"requestId"`
 	Identity     APIGatewayRequestIdentity `json:"identity"`
 	ResourcePath string                    `json:"resourcePath"`
+	Authorizer   map[string]interface{}    `json:"authorizer"`
 	HTTPMethod   string                    `json:"httpMethod"`
 	APIID        string                    `json:"apiId"` // The API Gateway rest API Id
 }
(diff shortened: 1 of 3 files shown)


events/apigw.go
```diff
@@ -27,14 +27,15 @@ type APIGatewayProxyResponse struct {
...
-	ResourcePath string                    `json:"resourcePath"`
-	HTTPMethod   string                    `json:"httpMethod"`
-	APIID        string                    `json:"apiId"` // The API Gateway rest API Id
+	AccountID    string                            `json:"accountId"`
+	ResourceID   string                            `json:"resourceId"`
+	Stage        string                            `json:"stage"`
+	RequestID    string                            `json:"requestId"`
+	Identity     APIGatewayRequestIdentity         `json:"identity"`
+	ResourcePath string                            `json:"resourcePath"`
+	Authorizer   APIGatewayCustomAuthorizerContext `json:"authorizer"`
```
bmoffatt: make this field `Authorizer map[string]interface{}` instead of changing `APIGatewayCustomAuthorizerContext`

deluan: It was my original code... :) I'll change it

deluan: Hey @bmoffatt, do you want to keep the old struct in the code?

bmoffatt: Leave it for now. Technically removing it would be a breaking change ;)

PR #5, opened 2018-01-16
Provide cloudwatch logs event, parse for content

jspc:
This pull request implements a cloudwatch logs event and provides tests and documentation strings.

Additionally will, when asked, parse the base64'd gzipped log data into a log entry.

PR #12, opened 2018-01-17
Cloudwatch Logs Helper functions

jspc:
## Note

This change is branched from https://github.com/jspc/aws-lambda-go/tree/cloudwatch_logs_events and, as such, includes commits from there. This will look far cleaner when/if https://github.com/aws/aws-lambda-go/pull/5 is merged

---

As per https://github.com/aws/aws-lambda-go/pull/5#issuecomment-358157294 in https://github.com/aws/aws-lambda-go/pull/5

These changes will allow for users, when parsing events, to do:

```golang
func Handler(ctx context.Context, input events.CloudwatchLogsEvent) (output interface{}, err error) {
        lData, err := input.AWSLogs.Parse()
        if err != nil {
                return
        }

// etc.
}
```

This replaces the need for the same boiler plate:

```golang
	data, err := base64.StdEncoding.DecodeString(cloudwatchLogsEevent.Data)
	if err != nil {
		return
	}

	zr, err := gzip.NewReader(bytes.NewBuffer(data))
	if err != nil {
		return
	}

	buf := &bytes.Buffer{}
	buf.ReadFrom(zr)

	output := map[string]interface{}{}

	err = json.Unmarshal(buf.Bytes(), &output)
	return

```


jspc:
## Addendum: 

I based this format on: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html

And from decoding the body from the default lambda test event for cloudwatch logs./

jspc:
@bmoffatt - were you able to get anything to discount the format from the docs? Or can we merge this in?

jspc:
@bmoffatt - good point, changed in https://github.com/aws/aws-lambda-go/pull/12/commits/f5a71e748b56dd653e829af1b9fea0f5e910c7a1

jspc:
@bmoffatt I don't really like the stuttering of the word `Log` but I've made the change anyway. ♻️

PR #16, opened 2018-01-18
Initial commit for handling of autoscalingEvents

ibrokethecloud:
The repo was missing ability to handle asg lifecycle events. 

Added functionality to handle autoscaling lifecycle events and action on them.

### DIFF
--- a/events/autoscaling.go
+++ b/events/autoscaling.go
@@ -6,38 +6,13 @@ import (
 
 // AutoScalingEvent struct is used to parse the json for auto scaling event types //
 type AutoScalingEvent struct {
-	Version    string                  `json:"version"`     // The version of event data
-	ID         string                  `json:"id"`          // The unique ID of the event
-	DetailType string                  `json:"detail-type"` //Details about event type
-	Source     string                  `json:"source"`      //Source of the event
-	AccountID  string                  `json:"account"`     //AccountId
-	Time       time.Time               `json:"time"`        //Event timestamp
-	Region     string                  `json:"region"`      //Region of event
-	Resources  []string                `json:"resources"`   //Information about resources impacted by event
-	Detail     AutoScalingEventDetails `json:"detail"`
-}
-
-// AutoScalingEventDetails struct is used to parse the nested event Details available within the Autoscaling event
-type AutoScalingEventDetails struct {
-	StatusCode           string                       `json:"StatusCode,omitempty"`           //Status code for Event
-	AutoScalingGroupName string                       `json:"AutoScalingGroupName,omitempty"` //AutoScalingGroup name
-	ActivityID           string                       `json:"ActivityId,omitempty"`           //ActivityId
-	Details              *AutoScalingEventDetailsInfo `json:"Details,omitempty"`              //Additional details about event such as AZ and SubnetId of instance
-	RequestID            string                       `json:"RequestId,omitempty"`            //RequestId of Request
-	StatusMessage        *string                      `json:"StatusMessage,omitempty"`        //StatusMessage
-	EndTime              *time.Time                   `json:"EndTime,omitempty"`              //EndTime of Event
-	EC2InstanceID        string                       `json:"EC2InstanceId,omitempty"`        //InstanceID of ec2 instance impacted by event
-	StartTime            *time.Time                   `json:"StartTime,omitempty"`            //StartTime of the event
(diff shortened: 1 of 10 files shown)


events/autoscaling.go
```diff
@@ -0,0 +1,42 @@
...
+	Region     string       `json:"region"`      //Region of event
+	Resources  []string     `json:"resources"`   //Information about resources impacted by event
+	Detail     EventDetails `json:"detail"`
+}
+
+// EventDetails struct is used to parse the nested event Details available within the Autoscaling event
+type EventDetails struct {
+	StatusCode           string      `json:"StatusCode,omitempty"`           //Status code for Event
+	AutoscalingGroupName string      `json:"AutoScalingGroupName,omitempty"` //AutoScalingGroup name
+	ActivityId           string      `json:"ActivityId,omitempty"`           //ActivityId
```
bmoffatt: change to `ActivityID`

events/autoscaling.go
```diff
@@ -0,0 +1,42 @@
...
+	Details              DetailsInfo `json:"Details,omitempty"`              //Additional details about event such as AZ and SubnetId of instance
+	RequestId            string      `json:"RequestId,omitempty"`            //RequestId of Request
```
bmoffatt: change to `RequestID`

events/autoscaling.go
```diff
@@ -0,0 +1,42 @@
...
+	StatusMessage        string      `json:"StatusMessage,omitempty"`        //StatusMEssage
+	EndTime              time.Time   `json:"EndTime,omitempty"`              //EndTime of Event
+	EC2InstanceId        string      `json:"EC2InstanceId,omitempty"`        //InstanceID of ec2 instance impacted by event
```
bmoffatt: change to `EC2InstanceID`

events/autoscaling.go
```diff
@@ -0,0 +1,42 @@
...
+	Source     string       `json:"source"`      //Source of the event
+	AccountID  string       `json:"account"`     //AccountId
+	Time       time.Time    `json:"time"`        //Event timestamp
+	Region     string       `json:"region"`      //Region of event
+	Resources  []string     `json:"resources"`   //Information about resources impacted by event
+	Detail     EventDetails `json:"detail"`
+}
+
+// EventDetails struct is used to parse the nested event Details available within the Autoscaling event
+type EventDetails struct {
```
bmoffatt: to avoid polluting the package with many simple type names, change this to something like `AutoScalingEventDetail`

events/autoscaling.go
```diff
@@ -0,0 +1,42 @@
...
+	StartTime            time.Time   `json:"StartTime,omitempty"`            //StartTime of the event
+	Cause                string      `json:"Cause,omitempty"`                //Cause of event
+	LifecycleActionToken string      `json:"LifecycleActionToken,omitempty"` //LifecycleActionToken details
+	LifecycleHookName    string      `json:"LifecycleHookName,omitempty"`    //LifecycleHookName
+	LifecycleTransition  string      `json:"LifecycleTransition,omitempty"`  //LifecycleTransition
+	NotificationMetadata string      `json:"NotificationMetadata,omitempty"` //Notification metadata
+}
+
+// DetailsInfo contains AZ and SubnetId for instance impacted by the autoscaling event //
+type DetailsInfo struct {
```
bmoffatt: to avoid polluting the package with many simple type names, change this to something like `AutoScalingEventDetailDetails`

ibrokethecloud:
Updated the code to have additional test cases for all samples available in the product documentation.

Have changed certain fields for AutoScalingEvents to pointer types to allow encoding to effectively handle empty fields which helps with the test cases.

bmoffatt:
Thanks for adding the other event samples.

After looking at the samples again, this approach creates a weird user experience. The `Details` member being the union all optional fields of all the different AutoScaling events means that a user should check against the empty string or nil before every use.

What do you think of having a separate AutoScaling event type for each of the samples in the documents? The alternative would be to only model the generic `CloudWatchEvent` with the `Detail` field as a `map[string]interface{}`

ibrokethecloud:
I am in favour of converinting the `Detail` field into a `map[string]interface{}` and adding additional event typed `Detail` structures to allow parsing the results based on requirements.

I will make the changes soon and update the request.

PR #22, opened 2018-01-20
GoDoc and Go Report Card badges

imdario:
It is a good practice to add these badges, alongside others like coverage :smile: They are handy.


dowlingw:
This is a prerequisite to listing this on `awesome-go` which this should definitely be listed on :)

bmoffatt:
Thanks for the additions!

PR #24, opened 2018-01-22
Appears as if result of `dep ensure` was modified. Missing projects tags.

pts-iankennedy:
Using https://github.com/gogradle/gogradle I noticed it would NPE searching for projects during dependency resolution. `go ensure` output should probably not be modified.


bmoffatt:
Looks like `dep ensure` wasn't run after adding [cmd/build-lambda-zip](https://github.com/aws/aws-lambda-go/tree/master/cmd/build-lambda-zip). I've checked in the new lock file, but it didn't contain the first project entry in your PR. Let us know if the issue you're seeing with gogradle is resolved.

pts-iankennedy:
Not resolved:

`:resolveBuildDependencies FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':resolveBuildDependencies'.
> Exception in resolution, message is:
  Cannot resolve dependency:github.com/github.com/aws: commit='LATEST_COMMIT', urls=[https://github.com/github.com/aws.git, git@github.com:github.com/aws.git]`


Requires:
`[[projects]]
  name = "github.com/aws/aws-lambda-go"`


The project name should be generated along with dep ensure.

PR #27, opened 2018-01-24
Use simple string concatenation

moorereason:


bmoffatt:
nice! removed an import!

PR #29, opened 2018-01-25
Add custom authorizer request and response

deluan:
This PR introduces the required structs for implementing an APIGW Custom Authorizer, as described in https://docs.aws.amazon.com/apigateway/latest/developerguide/use-custom-authorizer.html

### DIFF
--- a/events/apigw.go
+++ b/events/apigw.go
@@ -61,3 +61,30 @@ type APIGatewayCustomAuthorizerContext struct {
 	NumKey      *int    `json:"numKey,omitempty"`
 	BoolKey     *bool   `json:"boolKey,omitempty"`
 }
+
+// APIGatewayCustomAuthorizerRequest contains data coming in to a custom API Gateway authorizer function.
+type APIGatewayCustomAuthorizerRequest struct {
+	Type               string `json:"type"`
+	AuthorizationToken string `json:"authorizationToken"`
+	MethodArn          string `json:"methodArn"`
+}
+
+// APIGatewayCustomAuthorizerResponse represents the expected format of an API Gateway authorization response.
+type APIGatewayCustomAuthorizerResponse struct {
+	PrincipalID        string                           `json:"principalId"`
+	PolicyDocument     APIGatewayCustomAuthorizerPolicy `json:"policyDocument"`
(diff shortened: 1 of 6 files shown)


events/apigw.go
```diff
@@ -61,3 +61,30 @@ type APIGatewayCustomAuthorizerContext struct {
...
+}
+
+// APIGatewayCustomAuthorizerPolicy represents an IAM policy
+type APIGatewayCustomAuthorizerPolicy struct {
+	Version   string
+	Statement []IAMPolicyStatement
+}
+
+type IAMPolicyStatement struct {
+	Action   string
```
suruseas: I think the array is better for 'Action' and 'Resource'

jkaflik: I don't agree with you. I'm currently running with my own structs for APIGW Authorizer events and I am having the same implementation right here.

Action and Resource are multiplied on parent's level, so you define more than one policy statement.

jkaflik: That's right, I got same own implementation like in this PR, but according to the documentation, we can pass list of actions/resources here.

deluan: Yeah, I based my implementation on the JS example in the [Custom Authorizer](https://docs.aws.amazon.com/apigateway/latest/developerguide/use-custom-authorizer.html) docs, but looks like the fields should actually be sets/lists/arrays... I'll updte the code and also add an example authorizer, based on the JS example in the docs

suruseas: Thanks!

jackrk: Thanks for writing this!

Yep both Action and Resource should be arrays. It would be great to also add [Condition](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition.html) as a map. 

Would you mind submitting the same PR to this repo? https://github.com/awslabs/aws-apigateway-lambda-authorizer-blueprints

deluan: Sure, I'll make the modifications this weekend

bmoffatt:
Thanks for the PR!

deluan:
Hey @jackrk, I added this code to the blueprints repo here is the PR: https://github.com/awslabs/aws-apigateway-lambda-authorizer-blueprints/pull/18

Issue #20, opened 2018-01-19
Event for CloudWatch scheduled function execution


ravernkoh:
Looking for the equivalent of [this](https://github.com/apex/apex-go/blob/master/cloudwatch/cloudwatch.go), which is an struct mapping for a CloudWatch scheduled function execution event.

Am I completely missing something here?

mpgerlek:
+1

binary132:
Same

bmoffatt:
merged 5fea2b254568d4043f23d27471ae265cd33d3530

PR #31, opened 2018-01-25
Adding events/cloudwatch_scheduled

jbristow:
PR Addressing #20

Tested in my personal account with a handler with this class as the event type. 

If further tests are required, please let me know and I'll add them.

### DIFF
--- a/events/cloudwatch_scheduled.go
+++ b/events/cloudwatch_scheduled.go
@@ -1,0 +1,19 @@
+package events
+
+import (
+	"encoding/json"
+	"time"
+)
+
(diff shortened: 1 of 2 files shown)


bmoffatt:
looks good! I made a change to the name and renamed the files to `CloudWatchEvent`/`cloudwatch_events.go`. The structure here is generic to the outer structure of all CloudWatch Events https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/EventTypes.html

jbristow:
Oh! That’s good! I was laser focused on my current work and didn’t consider it might be more generic.

Thanks for the quick turnaround!

PR #33, opened 2018-01-28
update imports in s3 event readme

eddturtle:
fmt and context weren't imported and strings was never used. Also moved the closing parenthesis onto a new line.


bmoffatt:
Thanks!

PR #43, opened 2018-02-16
Added .travis.yml for CI

### LINKED ISSUE #10: Integrate repository with a CI system
Hi,

When sending PRs, it would be useful if the repository would be integrated with CI systems like CircleCI so that we can see if a PR fails or not. Thanks.

adamo57:
I chose to go with travis because it is what I am most familiar with, open to discussions around this implementation.

Fixes #10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,40 @@
+language: go
+
+go:
+  - 1.8
+  - 1.9
+  - tip 
+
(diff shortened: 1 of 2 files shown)


bmoffatt:
Thanks for the PR - I had trouble with the .travis.yml provided when I configured bmoffatt/aws-lambda-go, and ended up with 9b2186b

Issue #48, opened 2018-02-27
Support for AWS CodePipeline


brianterry:
Hello, there are no events for AWS CodePipeline jobs.  This would be great to have.

brianterry:
I have a PR open for this issue.  I was looking at the AWS docs, and noticed that their sample JSON includes "null".  https://docs.aws.amazon.com/codepipeline/latest/userguide/actions-invoke-lambda-function.html#actions-invoke-lambda-function-json-event-example  In the  unmarshal and marshal event, go converts this to " "

bmoffatt:
merged #53

PR #53, opened 2018-03-01
Codepipeline

brianterry:
Fixes issue #48

### DIFF
--- a/events/codepipeline_job.go
+++ b/events/codepipeline_job.go
@@ -1,0 +1,74 @@
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+package events
+
+// CodePipeline contains data from an event sent from AWS Codepipeline
+type CodePipeline struct {
+	CodePipelineJob CodePipelineJob `json:"CodePipeline.job"`
+}
+
+// CodePipelineJob represents a job from an AWS CodePipeline event
+type CodePipelineJob struct {
+	ID        string           `json:"id"`
+	AccountID string           `json:"accountId"`
+	Data      CodePipelineData `json:"data"`
+}
+
+// CodePipelineData represents a job from an AWS CodePipeline event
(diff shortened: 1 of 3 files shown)


events/codepipeline_job.go
```diff
@@ -0,0 +1,74 @@
...
+// CodePipelineConfiguration represents a configuration for an Action Configuration
+type CodePipelineConfiguration struct {
+	FunctionName   string `json:"FunctionName"`
+	UserParameters string `json:"UserParameters"`
+}
+
+// CodePipelineInputArtifact represents an input artifact
+type CodePipelineInputArtifact struct {
+	Location CodePipelineInputLocation `json:"location"`
+	Revision string                    `json:"revision"`
```
bmoffatt: regarding your comment on #48 
> In the unmarshal and marshal event, go converts this to " "

You can make the documented sample from https://docs.aws.amazon.com/codepipeline/latest/userguide/actions-invoke-lambda-function.html#actions-invoke-lambda-function-json-event-example work if you change this field's type to be pointer. `Revision *string`

brianterry: Changing `Revision string`  to `Revision *string` still results in Revision being converted to   `revision":" " ` if `"revision": null,` is given in the test doc. This causes tests to fail if using https://docs.aws.amazon.com/codepipeline/latest/userguide/actions-invoke-lambda-function.html#actions-invoke-lambda-function-json-event-example as a test doc.  Should I just replace `"revision": null,` with `"revision": "foo"` in the test doc?

bmoffatt: From https://play.golang.org/p/azCGae4c6jL, the type change ought to work.

bmoffatt:
Make the `string` -> `*string` change and updated the test data. Thanks for the contribution!

brianterry:
Woot!

PR #62, opened 2018-03-24
API Gateway custom authorizer request type structs

### LINKED ISSUE #61: API Gateway Custom Authorizer of type REQUEST not supported
[aws-lambda-go/events/apigw.go:66](https://github.com/aws/aws-lambda-go/blob/master/events/apigw.go#L66) contains structs for custom authorizers of type TOKEN. However, there is a second custom authorizer of type REQUEST for which the request and context are different and require different structs, currently missing.

See https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-custom-authorizer-input.html for request inputs.

Submitted #62 for code review.

thinkski:
An API Gateway Custom Authorizer supports two request types --
token and request. The existing structs are for the token type of
request. This commit adds the request type of request.

Fixes #61.

### DIFF
--- a/events/apigw.go
+++ b/events/apigw.go
@@ -53,6 +53,12 @@ type APIGatewayRequestIdentity struct {
 	User                          string `json:"user"`
 }
 
+// APIGatewayCustomAuthorizerRequestTypeRequestIdentity contains identity information for the request caller.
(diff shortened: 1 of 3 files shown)


bmoffatt:
Thanks for the contribution!

PR #67, opened 2018-04-05
 Adding structures for iot custom authorizer request/response 

ynori7:
Adding structures to fit to the documentation here: https://docs.aws.amazon.com/iot/latest/apireference/API_TestInvokeAuthorizer.html


LegNeato:
@bmoffatt any chance we can get this merged?

events/iot.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// IoTCustomAuthorizerResponse represents the expected format of an IoT device gateway authorization response.
+type IoTCustomAuthorizerResponse struct {
+	IsAuthenticated          bool                   `json:"isAuthenticated"`
+	PrincipalID              string                 `json:"principalId"`
+	DisconnectAfterInSeconds int32                  `json:"disconnectAfterInSeconds"`
+	RefreshAfterInSeconds    int32                  `json:"refreshAfterInSeconds"`
+	PolicyDocuments          []string               `json:"policyDocuments"`
+	Context                  map[string]interface{} `json:"context,omitempty"`
```
bmoffatt: I don't see a context field in the documentation. Is there another reference that contains this?

from https://docs.aws.amazon.com/iot/latest/apireference/API_TestInvokeAuthorizer.html#API_TestInvokeAuthorizer_ResponseElements
```
{
   "disconnectAfterInSeconds": number,
   "isAuthenticated": boolean,
   "policyDocuments": [ "string" ],
   "principalId": "string",
   "refreshAfterInSeconds": number
}
```

events/iot.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// IoTCustomAuthorizerPolicy represents an IAM policy. PolicyDocuments is an array of IoTCustomAuthorizerPolicy JSON strings
```
bmoffatt: minor, It may make sense to add or use an existing IAMPolicy type instead of redefining the envelope. 

A README would also help to show expected usage.

ynori7:
> @bmoffatt any chance we can get this merged?

Yeah, it looks like the contract changed a lot since I made these changes. I've updated everything now and addressed all the comments.

ynori7:
@bmoffatt Could you take another look?

.idea/aws-lambda-go.iml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bmoffatt, line 1: project files need to be removed :)

events/iot.go
```diff
@@ -0,0 +1,34 @@
...
+	TokenSignature     string          `json:"tokenSignature"`
+}
+
+type IoTHttpContext struct {
+	Headers     map[string]string `json:"headers,omitempty"`
+	QueryString string            `json:"queryString"`
+}
+
+type IoTMqttContext struct {
+	ClientId string `json:"clientId"`
```
bmoffatt, line 18: style: general field and  type naming conventions should consistently capitalize these initialisms 

```
Id -> ID
HTTP -> HTTP
Tls->TLS
Mqtt -> MQTT
```

bmoffatt (changes requested):

events/policy.go
```diff
@@ -0,0 +1,7 @@
+package events
```
bmoffatt, line 1: unused file?

bmoffatt: going to delete this one before merging, feel free to re-open if this one was needed for something!

bmoffatt (approved):

events/policy.go
ynori7: I think it was used before, but not anymore after resolving the most recent merge conflicts since it was added already somewhere else with a different name. Thanks for cleaning up!

rittneje:
@ynori7 @bmoffatt Unfortunately, I think you referenced the wrong request structure. The actual structure is documented here: https://docs.aws.amazon.com/iot/latest/developerguide/config-custom-auth.html I will file a bug.

I guess for some reason the IoT Core developers decided to make the test request totally different from the real request.

PR #69, opened 2018-04-08
Start() to allow customizable Interface to be passed as wrapper handler

yunspace:
I realize that the current design is to abstract away the inner workings of both `Function` and `lambdaHandler`, letting the user go through `Start()` entry point only.

However there are some valid scenarios when a user may wish to implement their own wrapper handlers to deal with the raw `[]byte` slice such as:

- writing a custom handler that is specific rather than generic, thus not having to rely on reflection.
- non JSON scenarios such as [APIGateway binary data](https://aws.amazon.com/about-aws/whats-new/2016/11/binary-data-now-supported-by-api-gateway/), [Protobuf](https://theburningmonk.com/2017/09/using-protocol-buffers-with-api-gateway-and-aws-lambda/) or something like LIBSVM, which would fall out of the default `encoding/json` imposed by `lambdaHandler`. 

The way I've refactored is to make `lambdaHandler` implement a interface called `LambdaHandler`. This way the current inner workings are still hidden, and users would pass through their custom `LamdbaHandler` via a new `StartWrapper()` entrypoint. 

An alternative would be to expose the function type of `lambdaHandler`. However, an interface seems more appropriate since
- `lambdaHandler` is already treated like an interface by calling the `Invoke()` method instead of being called as a function: https://github.com/aws/aws-lambda-go/blob/master/lambda/function.go#L59
- function implementation can have different signature to the `Invoke()` function, i.e. returning `interface{}` instead of `[]byte`: https://github.com/aws/aws-lambda-go/blob/master/lambda/handler.go#L13
- can be implemented using structs instead of functions

### DIFF
--- a/lambda/entry.go
+++ b/lambda/entry.go
@@ -38,23 +38,23 @@ import (
 // See https://golang.org/pkg/encoding/json/#Unmarshal for how deserialization behaves
 func Start(handler interface{}) {
 	wrappedHandler := newHandler(handler)
-	StartWrapper(wrappedHandler)
+	StartHandler(wrappedHandler)
 }
 
-// StartWrapper takes in a LambdaHandler wrapper interface which can be implemented either by a
+// StartHandler takes in a Handler wrapper interface which can be implemented either by a
 // custom function or a struct.
 //
-// LambdaHandler wrapper implementation requires a single "Invoke()" function:
+// Handler implementation requires a single "Invoke()" function:
 //
 //  func Invoke(context.Context, []byte) ([]byte, error)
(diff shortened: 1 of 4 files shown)


bmoffatt:
Looks good. I made a change to the interface and function name. Leaving the PR open for a little while to sleep on the naming of `lambda.StartHander/StartWrapper`.

yunspace:
Thanks @bmoffatt. Happy with the `lambda.StartHandler` and `Handler` naming. Seems to be inline with the `http.Handle()` and `http.HandleFunc()` style.

PR #71, opened 2018-04-11
lambdacontext: Add Deadline to the context

jmassara:
This change adds the deadline to the LambdaContext type. This will allow
users to know how much time is remaining for a given invocation.

Using `time.Until` on the deadline is similar to the
`get_remaining_time_in_millis` method for the Python context object. See
https://docs.aws.amazon.com/lambda/latest/dg/python-context-object.html.


bmoffatt:
the deadline is already being set on the `context.Context` -https://github.com/aws/aws-lambda-go/blob/master/lambda/function.go#L37-L38 

You can see an example of its use on https://docs.aws.amazon.com/lambda/latest/dg/go-programming-model-context.html

Issue #66, opened 2018-04-04
LexEvent - ResponseCard error


iliasgal:
I am not sure if this is a bug.

I am using a Lambda function as fulfilment to AWS Lex. I want to send a response using the code below:  

```
func scheduleMeeting(event events.LexEvent) (events.LexEvent, error) {

    //some code here

    response := events.LexEvent{
        DialogAction: &events.LexDialogAction{
            Type:             "Close",
            FulfillmentState: "Fulfilled",
            Message:          map[string]string{"contentType": "PlainText", "content": "Thank you!"},
        },
    }

    return response, nil
}
```
However, I am getting the following error:

> An error has occurred: Invalid Lambda Response: Received invalid response from Lambda: Can not deserialize value of type ContentType from String "": value not one of declared Enum instance names: http://application/vnd.amazonaws.card.generic at Source: {"sessionAttributes":null,"dialogAction":{"type":"Close","fulfillmentState":"Fulfilled","message":{"content":"Thank you!","contentType":"PlainText"},"intentName":"","slots":null,"slotToElicit":"","responseCard":{"version":0,"contentType":"","genericAttachments":null}}}; line: 1, column: 239

This `ContentType` field and the Enum `application/vnd.amazonaws.card.generic` are on the `ResponseCard` field, which is optional according to the [documentation](https://docs.aws.amazon.com/lex/latest/dg/lambda-input-response-format.html#using-lambda-response-format).

asuresh8:
Hmm that is strange, and if you add the the `ContentType: application/vnd.amazonaws.card.generic` in the `ResponseCard` object does the error go away?

iliasgal:
I added the `ResponseCard` object.

```
var buttons []map[string]string
buttons = append(buttons, map[string]string{"text": "test", "value": "test"})

var attachements []events.Attachment
attachements = append(attachements, events.Attachment{
	Title:    "test",
	SubTitle: "test subtitle",
        ImageURL: "https://testurl.com",
	Buttons:  buttons,
})

response := events.LexEvent{
	DialogAction: &events.LexDialogAction{
		Type:             "Close",
		FulfillmentState: "Fulfilled",
		Message:          map[string]string{"contentType": "PlainText", "content": message},
		ResponseCard: events.LexResponseCard{
			Version:            1,
			ContentType:        "application/vnd.amazonaws.card.generic",
			GenericAttachments: attachements,
		},
	},
}
```

And the new error is:

[...]

PR #75, opened 2018-04-25
Add omitempty for lex struct fields

iliasgal:
Fix for issue #66 

`Omitempty` has been added for the lex.go structs. Some fields are required and technically do not need `omitempty`, but we can just let AWS handle that and return an error message if a user does not add a required field.


bmoffatt:
Thanks!

PR #78, opened 2018-04-28
"OutPutArtifacts" -> "OutputArtifacts"

LegNeato:


LegNeato:
Such is life with hand-generated types, sigh.

PR #93, opened 2018-07-10
Modified README.md

chorinator:
Added support for macos packaging. 
Sometimes the macos zip command gives an error when running the lambda on aws

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,16 +25,16 @@ To learn more about writing AWS Lambda functions in Go, go to [the offical docum
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


leepa:
Could you be more specific around how sometimes the macOS zip command gives an error when running the lambda? You've also flattened out the example code, looking at the diff and therefore that would need fixing.

bmoffatt:
The `build-lambda-zip` tool isn't currently flexible enough to recommend as the default zip tool for function developers. Feel free to open a separate issue regarding using the default zip installation on macOS.

PR #97, opened 2018-07-24
Update Kinesis event readme

atsushi-ishibashi:
> The Data attribute in the Lambda record is Base64 encoded and compressed with the gzip format.
https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html

You can receive payload which already is Base64 decoded so gzip decompression will be needed.


bmoffatt:
the requirement to use gzip decompression is unique to CloudWatch Subscription Filters, so this particular readme wouldn't be the right place to document it. Do feel free to update the PR to add an additional section for CloudWatch Subscription Filters.

Issue #96, opened 2018-07-20
build-lambda-zip command only takes one argument


Jeff-Hanson:
The build-lambda-zip utility only takes one argument to include in the resultant zip file. There are cases where it is necessary to include additional files beyond the executable, for example we want to include a cert file that can be read by the lambda. There is no way to include that file in the zip that is built by build-lambda-zip.exe. An example of the desired behavior would be:

build-lambda-zip.exe -o test.zip test test.crt 

which would result in test.zip containing both test and test.crt.

leepa:
You can add files to file using the conventional programs. All the build-lambda-zip executable does is mark the lambda itself as an executable file. This is a particular hassle on Windows, hence its existence.

Jeff-Hanson:
True that can be done but on Windows I think that means either using a 3rd party tool or PowerShell in Windows. That means dealing with another tool and keeping it up to date. Which is not ideal.

leepa:
If the one-liner Powershell, which is the default command processor in newer builds on Windows, isn't okay then maybe embedding the files in your Go exe using something like https://github.com/gobuffalo/packr - that'll give something more Go native. There's no clean option that doesn't involve more steps though.

For reference though (for others looking at this issue):

Compress-Archive -Path test.crt -Update -DestinationPath test.zip

Personally, I'm reluctant to turn build-lambda-zip into 'yet another generic zip tool' - consider putting in a pull request as it might represent the argument better.

Jeff-Hanson:
Looks like powershell mucks with the permissions when using update so I don't think that will work. Packr does seem to work and could be an option.  I put in a pull request with the changes I used to accomplish this.

PR #98, opened 2018-07-25
Multiple files windows zip

Jeff-Hanson:
In response to [issue 14 ](https://github.com/aws/aws-lambda-go/issues/96) this is an enhancement to support adding multiple files to a zip on windows using the build-lambda-zip tool.

### DIFF
--- a/cmd/build-lambda-zip/main.go
+++ b/cmd/build-lambda-zip/main.go
@@ -14,12 +14,12 @@ import (
 func main() {
 	app := cli.NewApp()
 	app.Name = "build-lambda-zip"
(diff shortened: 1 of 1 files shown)


leepa:
Please don't include binaries in the repository, might be worst squashing/force pushing as well as you'll have to force push anyway.

Jeff-Hanson:
OOPS, sorry about the binary. Should be removed now.

bmoffatt:
I'm happy with this approach. This tool may now need a v2 revision when adding additional flags.

ioannist:
Tried including a template file, but I am getting an error.

```
go build -o ..\dist\main main.go welcome.gohtml
named files must be .go files: welcome.gohtml
```

PR #102, opened 2018-08-04
Add dmarc policy to SES event

jmoney8080:
The SES Receipt is missing the attribute for dmarcPolicy of the inbound email.


bmoffatt:
aligns with https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-notifications-contents.html#receiving-email-notifications-contents-receipt-object 

Thanks!

PR #103, opened 2018-08-04
Add disposition type for controlling inbound

jmoney8080:
To control inbound email, a JSON string of the form `{disposition: <control>}` needs to be returned.  For instance if you want to stop the entire rule set you must return `{disposition: STOP_RULE_SET}` and inbound will stop there at the lambda.  Adding the type to the SES events as this was missing.  I've had to copy this type around to a few internal project and thought it best to centralize.  Let me know if there's a better location for this type.

### DIFF
--- a/events/ses.go
+++ b/events/ses.go
@@ -63,3 +63,8 @@ type SimpleEmailReceiptAction struct {
 type SimpleEmailVerdict struct {
 	Status string `json:"status"`
 }
+
+// SimpleEmailDisposition disposition return for SES to control rule functions
+type SimpleEmailDisposition struct {
+	Disposition string `json:"disposition"`
+}
--- a/events/ses.go
+++ b/events/ses.go
@@ -64,7 +64,19 @@ type SimpleEmailVerdict struct {
 	Status string `json:"status"`
 }
 
+// SimpleEmailDispositionValue enumeration representing the dispostition value for SES
+type SimpleEmailDispositionValue string
+
+const (
+	// SimpleEmailContinue represents the CONTINUE disposition which tells the SES Rule Set to continue to the next rule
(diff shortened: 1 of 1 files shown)


bmoffatt:
The location is fine. The `events` package gets used for modeling response types as well as input types.

per: https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-action-lambda.html
> and the second argument is a disposition property that is set to either STOP_RULE, STOP_RULE_SET, or CONTINUE. If the second argument is null or does not have a valid disposition property, the mail flow continues and further actions and rules are processed, which is the same as with CONTINUE. 

Since it appears that there is a known set of possible values, the type could look something like:
```
// SimpleEmailDisposition disposition return for SES to control rule functions
type SimpleEmailDispositionValue *string
const (
	SimpleEmailContinue    SimpleEmailDispositionValue = "CONTINUE";
	SimpleEmailStopRule    SimpleEmailDispositionValue = "STOP_RULE";
	SimpleEmailStopRuleSet SimpleEmailDispositionValue = "STOP_RULE_SET";
)
type SimpleEmailDisposition struct {
	Disposition SimpleEmailDispositionValue `json:"disposition"`
}
```

jmoney8080:
@bmoffatt I have adde the changes.  I had to make the type of the disposition value a string instead of a pointer to a string.  This is because you cannot assign literals to the const values as pointers.  golang doesn't let you do that.  I'd have to make separate const strings and then assign the pointers to the disposition value types, but that seemed a little redundant given that since these are already an enumeration there's not memory savings by just sending around pointers everywhere.

bmoffatt:
looks good to me. Thanks!

Issue #107, opened 2018-08-19
cfn wrapper for Cloudformation does not report correct error to aws


vmadura:
When an error is returned by the lambda wrapped by aws cfn wrapper; the PhysicalResourceID is not included. It was modified recently to include all request types: https://github.com/aws/aws-lambda-go/pull/76

However, AWS Docs mention PhysicalResourceID is always required. Therefore Cloudformation incorrectly reports the error "Invalid Physical Resource ID" instead of the actual error during resource creation.

**Steps to replicate:** 
Modified the Simple Test Lambda to include an error condition:
```
package main

import (
  "context"
  "fmt"
  "errors"
  "github.com/aws/aws-lambda-go/cfn"
  "github.com/aws/aws-lambda-go/lambda"
)

func echoResource(ctx context.Context, event cfn.Event) (physicalResourceID string, data map[string]interface{}, err error) {
    v, _ := event.ResourceProperties["Echo"].(string)
    if event.RequestType == "Create" {
      if v == "ERROR" {
	err = errors.New("My Custom Resource Creation Failed")
	fmt.Errorf("Error occurred")
	return 
      } 
    }
    data = map[string]interface{} {
        "Echo": v,
    }

    return
}

func main() {
	lambda.Start(cfn.LambdaWrap(echoResource))
}
```

**Created simple Cloudformation template to create custom resource:**
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: AWS Test

Resources:
  TestResource:
    Type: "Custom::TestResource"
    Properties:
      Echo: ERROR
      ServiceToken: arn:aws:lambda:us-west-1:xxxxx:function:cfn-test
```

**Observed that the error reported during Stack Creation is incorrect:** 
![image](https://user-images.githubusercontent.com/8883916/44306235-65bdba00-a33f-11e8-9507-ee9f83d8baed.png)

The error message reported must be : `My Custom Resource Creation Failed`

kh01:
@vmadura: I have a question that is not related to issue 107
I ran into the issue where if I set err = errors.New("My Custom Resource Creation Failed") for the event "event.RequestType == "Update", the cloudformation template will go into the failed state and cannot be update or delete. Do you know how to handle this situation?

Thank in advanced

bmoffatt:
reopening per #241

rix0rrr:
[...]

PR #108, opened 2018-08-19
fixes error messaging - physical resource id is always required

vmadura:
Fixes https://github.com/aws/aws-lambda-go/issues/107

Tested using same steps in the reported issue and confirmed that correct error message is being reported:
image


bmoffatt:
Thanks for the fix!

PR #110, opened 2018-08-23
DynamoDB Lambda Trigger

sesheffield:
All the other events that trigger a lambda function nicely connect with the go aws sdk, except for Dynamodb. 
Updating 'Keys', 'NewImage', and 'OldImage' in 'DynamoDBStreamRecord' struct to allow for 'dynamodbattribute.UnmarshalMap()' functionality when triggering a lambda function with dynambodb.


leepa:
This change heavily bloats dependencies. Also if you are vendoring things, best not to create a dependency loop and vendor the same library. You also should not run tools on things in the vendor folder.

sesheffield:
@leepa I removed any changes to the vendor file to prevent any import cycles and remove the 3000+ file bloat

leepa:
@sesheffield you still have aws-lambda-go vendored inside aws-lambda-go. At least at the point of writing this.

sesheffield:
@leepa Sorry I did not realize I had vendors in the PR still.

events/testdata/dynamodb-event.json
```diff
@@ -14,27 +14,81 @@
         "ApproximateCreationDateTime": 1480642020,
         "Keys": {
           "val": {
-            "S": "data"
+             "B": null,
```
harrisonhjones: Looks (at least in my browser) that the alignment is still a bit off here.

events/attributevalue.go
```diff
@@ -18,6 +18,71 @@ type DynamoDBAttributeValue struct {
 	dataType DynamoDBDataType
 }
 
+// Replicating this from aws sdk to prevent dependency bloating / potential dependency import cycles
```
harrisonhjones: Ideally this comment would still follow standard godoc pattern (ie, `An AttributeValues is ...`). This might also be a good place to explain where it's replicated from (the AWS SDK) and how to update it in the future.

events/attributevalue.go
```diff
@@ -18,6 +18,73 @@ type DynamoDBAttributeValue struct {
 	dataType DynamoDBDataType
 }
 
+// AttributeValue is the data for an attribute that has a name-value pair. The name is the data type, and the value is the actual data.
+// This is being replicated from "https://docs.aws.amazon.com/sdk-for-go/api/service/dynamodb/#AttributeValue"(AWS SDK) to prevent dependency bloating or potential dependency import cycles.
```
bmoffatt: https://github.com/aws/aws-lambda-go/blob/d56e45b306530c767d46373096c2000d3478284f/events/attributevalue.go#L21-L23

How does replicating the type enable it's use with (dynamodbattribute.UnmarshalMap)[https://github.com/aws/aws-sdk-go/blob/f26268b77ea9619d95b1d04b6f3e3c5a03b61110/service/dynamodb/dynamodbattribute/decode.go#L86]? A test case would be helpful in understanding the usage.

events/dynamodb.go
```diff
@@ -72,13 +72,13 @@ type DynamoDBStreamRecord struct {
 	ApproximateCreationDateTime SecondsEpochTime `json:"ApproximateCreationDateTime,omitempty"`
 
 	// The primary key attribute(s) for the DynamoDB item that was modified.
-	Keys map[string]DynamoDBAttributeValue `json:"Keys,omitempty"`
```
bmoffatt: This breaks backwards compatibility for existing consumers.

events/attributevalue.go
WIZARDISHUNGRY: Seconded: I'd really like to be able to use `UnmarshalMap` in my lambda handlers

harrisonhjones: What if we wrote, in perhaps another package to reduce the bloat of this package, a translator that translated DynamoDBAttributeValue to a dynamodb.*AttributeValue ? Then, if you needed such functionality, you could import that package and use it.

WIZARDISHUNGRY: In an  aid to those googling, there's some discussion of this here: https://stackoverflow.com/questions/49129534/unmarshal-mapstringdynamodbattributevalue-into-a-struct/50164289#50164289?newreg=c2e95160086c4a21a406a2e6ed0d15d9

bmoffatt:
closing stale PR - there's a few issues open already tracking the conversation for what else could be done here

PR #116, opened 2018-09-08
README for AppSync

atsushi-ishibashi:
closed: #115


bmoffatt:
Thanks for adding a sample!

Issue #118, opened 2018-09-13
Missing notifcationType in SimpleEmailService


kaihendry:
https://github.com/aws/aws-lambda-go/blob/master/events/ses.go#L16

Put https://s.natalian.org/2018-09-13/ses.json into https://mholt.github.io/json-to-go/

kaihendry:
Furthermore https://github.com/aws/aws-lambda-go/blob/master/events/ses.go#L58-L62 is missing details like topicArn, bucketName, objectKey et al

bmoffatt:
resolved by #211

PR #119, opened 2018-09-13
Closes https://github.com/aws/aws-lambda-go/issues/118

kaihendry:
I am unfamiliar of tools to make sure the mapping between the JSON and struct is 100%. So it was a manual process.


bmoffatt:
resolved in #211

PR #120, opened 2018-09-19
Fix Lex event slots type

xfrr:
If you set Slots as "map[string]string", when you send them delegating the action, amazon lex suppose that the next slots have been filled and goes directly to the "FulfillmentCodeHook". So, the type of the slots should be "map[string]interface{}" to send back the slots values as (nil), not empty string ("").

### DIFF
--- a/events/lex.go
+++ b/events/lex.go
@@ -41,7 +41,7 @@ type LexDialogAction struct {
 	ResponseCard     *LexResponseCard  `json:"responseCard,omitempty"`
 }
 
-type Slots map[string]interface{}
+type Slots map[string]*string
 
 type LexResponseCard struct {
 	Version            int64        `json:"version,omitempty"`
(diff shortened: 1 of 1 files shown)


events/lex.go
```diff
@@ -41,7 +41,7 @@ type LexDialogAction struct {
 	ResponseCard     *LexResponseCard  `json:"responseCard,omitempty"`
 }
 
-type Slots map[string]string
+type Slots map[string]interface{}
```
bmoffatt: Do I understand right, this is to support  explicit null slots? ex: `"slots": { "slot-name": null }`. Could this instead be `map[string]*string`? It'd help to see the test sample expanded, and an AWS documentation reference.

xfrr: You are right, according to the AWS Reference should be a string pointer.
Btw, this is because when I delegate the next action of the intent to Amazon, if the next required slot is empty string, Amazon avoid it, like the slot have been already filled.
Thanks.

bmoffatt:
Thanks for the contribution!

PR #134, opened 2018-10-22
Make newhandler public

willnewrelic:
This change makes it much easier to write "middleware" for instrumentation, analytics, etc.

If the consumer registered their handler like this:
```go
func main() {
	lambda.Start(myHandler)
}
```
They may now use middleware like this:

```go
func addMiddleware(fn interface{}) lambda.Handler {
	h := lambda.NewHandler(fn)
	// h could easily be wrapped here to provide instrumentation here
	return h
}

func main() {
	lambda.StartHandler(addMiddleware(myHandler))
}
```

If `newHandler` is not public, writing middleware is real tough since the middleware layer must do reflection on the flexible `interface{}` function.

Thanks for your consideration!

### DIFF
--- a/lambda/entry.go
+++ b/lambda/entry.go
@@ -37,7 +37,7 @@ import (
 // Where "TIn" and "TOut" are types compatible with the "encoding/json" standard library.
 // See https://golang.org/pkg/encoding/json/#Unmarshal for how deserialization behaves
 func Start(handler interface{}) {
-	wrappedHandler := newHandler(handler)
+	wrappedHandler := NewHandler(handler)
 	StartHandler(wrappedHandler)
 }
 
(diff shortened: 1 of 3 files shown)


bmoffatt:
Looks good to me. When writing middleware packages, you might also define a function of the form 
`func AddMiddleware(next lambda.Handler) lambda.Handler` to support other types that implement the `lambda.Handler` interface, like what was described by #69.

willnewrelic:
@bmoffatt   Thanks.

Before merging this, let me open another PR today to add some more perspective on my middleware aims.

willnewrelic:
@bmoffatt   Thanks.

It would be nice to give middleware access to the request/response events.  I'm not sure what the best way to do that is, but I tried something here:

https://github.com/willnewrelic/aws-lambda-go/pull/1

What do you think?

PR #135, opened 2018-10-23
Add new multi value properties for Headers and Query

IainCole:
API Gateway just added these properties so this enables people to use them.


IainCole:
I shall write some tests soon

piotrkubisa:
I created a #136 which fixes an issue with unit tests.

bmoffatt:
Resolved by #136

PR #141, opened 2018-11-06
Introduce HandlerTrace callbacks

willnewrelic:
Hi

Now that `NewHandler` is public, it is possible to write "middleware" instrumentation that wraps the handler.  eg.  `lambda.Start(myHandler)`  becomes `lambda.StartHandler(addMiddleware(myHandler))`.  Fantastic!

Unfortunately, since the internal `lambdaHandler` does the JSON serialization and deserialization, the middleware doesn't have access to the request and response events. This is unfortunate because middleware instrumentation would really love to record request and response fields (such as the request event source ARN).

Consumers could be asked to modify their handlers to add callbacks, but it sure would be nice if the middleware could access the request and response automatically!!

Thinking through this problem, I was reminded of the [httptrace](https://golang.org/pkg/net/http/httptrace/) package which allows people to add http instrumentation by adding callbacks to the context. That's the idea presented here!

Example use:
```go
func myHandler(ctx context.Context, x int) (int, error) {
	return 456, nil
}

type handlerFunc func(ctx context.Context, payload []byte) ([]byte, error)

func (h handlerFunc) Invoke(ctx context.Context, payload []byte) ([]byte, error) {
	return h(ctx, payload)
}

// introduceMiddleware would exist in a third party package
func introduceMiddleware(originalHandler interface{}) lambda.Handler {
	handler := lambda.NewHandler(originalHandler)
	return handlerFunc(func(ctx context.Context, payload []byte) ([]byte, error) {
		ctx = lambda.WithHandlerTrace(ctx, lambda.HandlerTrace{
			RequestEvent: func(c context.Context, e interface{}) {
				fmt.Println("request", e)
			},
			ResponseEvent: func(c context.Context, e interface{}) {
				fmt.Println("response", e)
			},
		})
		return handler.Invoke(ctx, payload)
	})
}

func main() {
	lambda.StartHandler(introduceMiddleware(myHandler))
}
```

Thanks for your consideration!

### DIFF
--- a/lambda/handlertrace/trace.go
+++ b/lambda/handlertrace/trace.go
@@ -1,0 +1,45 @@
+// Package handlertrace allows middleware authors using lambda.NewHandler to
+// instrument request and response events.
+package handlertrace
+
+import (
+	"context"
+)
+
+// HandlerTrace allows handlers which wrap the return value of lambda.NewHandler
+// to access to the request and response events.
+type HandlerTrace struct {
+	RequestEvent  func(context.Context, interface{})
+	ResponseEvent func(context.Context, interface{})
+}
+
+func callbackCompose(f1, f2 func(context.Context, interface{})) func(context.Context, interface{}) {
+	return func(ctx context.Context, event interface{}) {
+		if nil != f1 {
+			f1(ctx, event)
+		}
+		if nil != f2 {
+			f2(ctx, event)
+		}
+	}
+}
+
+type handlerTraceKey struct{}
+
+// WithHandlerTrace adds callbacks to the provided context which allows handlers
+// which wrap the return value of lambda.NewHandler to access to the request and
+// response events.
+func WithHandlerTrace(ctx context.Context, trace HandlerTrace) context.Context {
+	existing := ContextHandlerTrace(ctx)
+	return context.WithValue(ctx, handlerTraceKey{}, HandlerTrace{
+		RequestEvent:  callbackCompose(existing.RequestEvent, trace.RequestEvent),
+		ResponseEvent: callbackCompose(existing.ResponseEvent, trace.ResponseEvent),
+	})
+}
+
+// ContextHandlerTrace returns the HandlerTrace associated with the provided
+// context.
+func ContextHandlerTrace(ctx context.Context) HandlerTrace {
+	trace, _ := ctx.Value(handlerTraceKey{}).(HandlerTrace)
+	return trace
+}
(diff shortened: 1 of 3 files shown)


willnewrelic:
Extra thoughts!

The context parameter to the callbacks might not be necessary (people can put state in another functions), but it seems like a convenient thing to have.

I chose functions (over interface values) for the callbacks because they were easier to compose.

willnewrelic:
This PR has been rebased on the latest master.  It is ready to go!

willnewrelic:
Hi @bmoffatt 

Great!!

I have rebased this PR and moved the new functions into a new `handlertrace` package.  Ready for your review!

I don't know what the best middleware story would be, but I think this is a good forward step!

I expect that only middleware authors are going to be using `lambda.NewHandler`, so that could be moved into `handlertrace` or some other internal package.  I have not done that to keep this PR focused and avoid breaking anything for consumers already using `lambda.NewHandler`.

lambda/handlertrace/trace.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+	}
+}
+
+type handlerTraceKey struct{}
+
+// WithHandlerTrace adds callbacks to the provided context which allows handlers
+// which wrap the return value of lambda.NewHandler to access to the request and
+// response events.
+func WithHandlerTrace(ctx context.Context, trace HandlerTrace) context.Context {
```
bmoffatt: rename as `NewContext`

lambda/handlertrace/trace.go
```diff
@@ -0,0 +1,45 @@
...
+	existing := ContextHandlerTrace(ctx)
+	return context.WithValue(ctx, handlerTraceKey{}, HandlerTrace{
+		RequestEvent:  callbackCompose(existing.RequestEvent, trace.RequestEvent),
+		ResponseEvent: callbackCompose(existing.ResponseEvent, trace.ResponseEvent),
+	})
+}
+
+// ContextHandlerTrace returns the HandlerTrace associated with the provided
+// context.
+func ContextHandlerTrace(ctx context.Context) HandlerTrace {
```
bmoffatt: rename as `FromContext`

lambda/handlertrace/trace.go
willnewrelic: Done.

lambda/handlertrace/trace.go
willnewrelic: Done.

bmoffatt:
Thanks! This'll get included in a release tag soon as well.

PR #143, opened 2018-11-08
Pass handler func []byte if the arg is interface{}

zbintliff:
When you pass a `func(ctx contex.Context, i interface{}) error{}` to `lambda.Start()` we
ran into this issue.  Since the wrapper took
`interface{}` as the second argument this would try to Unmarshal to it.

In `json.Unmarshal` if its an `interface{}` it unmarshals to
`map[string]interface`. But usually if its json you don't want a direct
transalation. For example `{"foo":"bar"}` turns into
`map[string]interface{}{"foo":"bar"}` even nesting all the way down.

This obviously changes the behavior (breaking even?) but we couldn't think of another way to let us get the raw `[]byte` back

### DIFF
--- a/lambda/handler.go
+++ b/lambda/handler.go
@@ -104,11 +104,18 @@ func NewHandler(handlerFunc interface{}) Handler {
 			eventType := handlerType.In(handlerType.NumIn() - 1)
 			event := reflect.New(eventType)
 
-			if err := json.Unmarshal(payload, event.Interface()); err != nil {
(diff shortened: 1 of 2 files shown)


zbintliff:
Adding on to this. If we define the second argument to be `[]byte` then it still fails :/

zbintliff:
Uh, long day and I"m an idiot. I see we can just use our how handler and deal with `[]byte`

PR #151, opened 2018-12-10
Fix field name in AppSync example 

aripalo:
Current `README_AppSync.md` file has wrong field name which causes `event.AppSyncOperation undefined (type events.AppSyncResolverTemplate has no field or method AppSyncOperation)` error when trying to run that sample code. The actual field name is `Operation` as seen in https://github.com/aws/aws-lambda-go/blob/master/events/appsync.go#L8.


aripalo:
Closing as duplicate of https://github.com/aws/aws-lambda-go/pull/137

PR #136, opened 2018-10-24
Add support for multi-value parameters in API Gateway proxy events

piotrkubisa:
Recently I was digging more into Lambda environment and I found new fields which are not defined in `aws-lambda-go` struct for API Gateway Proxy Request event:

- `events.APIGatewayProxyRequest.MultiValueHeaders`.
- `events.APIGatewayProxyRequest.MultiValueQueryStringParameters`.
- `events.APIGatewayRequestIdentity.AccessKey` (populated whether doing an authorization using AWS credentials).

https://aws.amazon.com/about-aws/whats-new/2018/10/amazon-api-gateway-adds-support-for-multi-parameters/

Maintainers: do we need any extra unit test for these new fields?

PR #155, opened 2019-01-04
Add multi-value headers to the API Gateway response

piotrkubisa:
I noticed `multiValueHeaders` is missing in the API Gateway response event.

- Relates to #136.
- https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format


bmoffatt:
Thanks!

Issue #154, opened 2019-01-04
json: cannot unmarshal string into Go struct field APIGatewayProxyRequestContext.authorizer


Yenchu:
When integrating with api gateway WebSocket API, got error in CloudWatch log:
`
json: cannot unmarshal string into Go struct field APIGatewayProxyRequestContext.authorizer of type map[string]interface {}: UnmarshalTypeError null
`

To log requestContext info, the field **authorizer** is empty:
`
requestContext:map[stage:... authorizer: messageId:<nil>...
`

piotrkubisa:
The type definition is a bit different for a Websocket events - for an instance (I might don't recall correctly) there are at least two fields which are missing in the `requestContext`:

```ts
connectionId: string
connectedAt: number
```

https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api-logging.html

That said, `aws-lambda-go` repository lacks with the struct for an API Gateway usage with a websocket. It would be great if someone could copy and paste the raw event payload here or make an PR and add it into the test data directory and create new struct with a corresponding `json` tags definition.

Dasio:
These are extra fields in APIGatewayProxyRequestContext
```go
	ConnectedAt        int64       `json:"connectedAt"`
	ConnectionID       string      `json:"connectionId"`
	DomainName         string      `json:"domainName"`
	Error              string      `json:"error"`
	EventType          string      `json:"eventType"`
	ExtendedRequestId  string      `json:"extendedRequestId"`
	IntegrationLatency string      `json:"integrationLatency"`
	MessageDirection   string      `json:"messageDirection"`
	// IDK type of this
	MessageID          interface{} `json:"messageId"`
	RequestTime        string      `json:"requestTime"`
	RequestTimeEpoch   int64       `json:"requestTime"`
	RouteKey           string      `json:"routeKey"`
	Status             string      `json:"status"`
```
But in WS event it is
```go
	Authorizer   string                      `json:"authorizer"`
```
Now it's 
```go
	Authorizer   map[string]interface{}    `json:"authorizer"`
```

jozanza:
[...]

Issue #153, opened 2019-01-02
ALB sample request payload has different Host header


piotrkubisa:
In the example payload in the `alb-lambda-target-request.json` file there is different value for a `Host` header (take a look at `multiValueHeaders` and `headers`):

https://github.com/aws/aws-lambda-go/blob/1be8070badf49117d0011e7267eda9b99f708604/events/testdata/alb-lambda-target-request.json#L20

https://github.com/aws/aws-lambda-go/blob/1be8070badf49117d0011e7267eda9b99f708604/events/testdata/alb-lambda-target-request.json#L39

PR #159, opened 2019-01-11
ApiGw websocket events

spezam:
Support for ApiGw Websocket event.

fix: https://github.com/aws/aws-lambda-go/issues/154


Dasio:
When WS Api gateway is using Authorize lambda function then field `Authorizer` in `APIGatewayWebsocketProxyRequestContext` is not string but map[string]interface{}

spezam:
@Dasio doesn't sound very consistent.

How should we call the struct when used for Authorize? Any example from other Lambda SDKs?

bmoffatt:
Was the event data in this PR and #153 generated from ex: docs, an example project? I'd like to play with the apparent inconsistencies.

PR #162, opened 2019-01-31
Refactor lambda/handler.go

willnewrelic:
This PR has no semantic impact.  It is purely an internals refactor.  I hope you think this makes the code a little cleaner!

* Before this PR, the functionality of the handler returned by `NewHandler` was split between the anonymous function within `NewHandler` and the `lambdaHandler.Invoke` method.  This was hard to follow.

* Before this PR, the request event type (`handlerType.In(handlerType.NumIn() - 1)`) was calculated on each invocation.  It is now calculated once.

I hope you enjoy!  I believe the risk is very low due to your great unit test coverage!


bmoffatt:
closing stale PR

PR #163, opened 2019-01-31
Add S3 test event

LegNeato:
As seen in https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html.

I didn't write a test as they are set up for a 1-1 mapping between event and service. Happy to add one, but it will take some test refactoring and for this simple struct that is never used in production I thought it likely wasn't worth it.

### DIFF
--- a/events/s3.go
+++ b/events/s3.go
@@ -51,3 +51,11 @@ type S3Object struct {
 	ETag          string `json:"eTag"`
 	Sequencer     string `json:"sequencer"`
 }
+
+type S3TestEvent struct {
+	Service   string    `json:"Service"`
+	Event     string    `json:"Event"`
+	Time      time.Time `json:"Time"`
+	RequestID string    `json:"RequestId"`
+	HostID    string    `json:"HostId"`
+}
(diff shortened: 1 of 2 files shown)


events/s3.go
```diff
@@ -51,3 +51,11 @@ type S3Object struct {
 	ETag          string `json:"eTag"`
 	Sequencer     string `json:"sequencer"`
 }
+
+type S3TestEvent struct {
```
bmoffatt: Missing `Bucket`

sample json from an sqs message
```
{"Service":"Amazon S3","Event":"s3:TestEvent","Time":"2019-02-04T19:34:46.985Z","Bucket":"bmoffatt","RequestId":"7BA1940DC6AF888B","HostId":"q1YDbiaMjllP0m+Lcy6cKKgxNrMLFJ9zCrZUFBqHGTG++0nXvnTDIGC7q2/QPAsJg86E8gI7y9U="}
```

bmoffatt:
I didn't see S3 send the test event to my Lambda function in testing, but could for SQS 🤷‍♂️

events/s3.go
LegNeato: Whoops, thanks for fixing!

LegNeato:
Right, but the SQS event appears to include this test event so it would be good to just have the def in the library.

PR #168, opened 2019-02-14
Add events for Cognito custom authentication challenge

mxpv:
This PR adds events for Cognito's custom authentication - https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-challenge.html


events/cognito.go
```diff
@@ -124,3 +124,72 @@ type GroupConfiguration struct {
...
+	CognitoEventUserPoolsHeader
+	Request  CognitoEventUserPoolsCreateAuthChallengeRequest  `json:"request"`
+	Response CognitoEventUserPoolsCreateAuthChallengeResponse `json:"response"`
+}
+
+// CognitoEventUserPoolsVerifyAuthChallengeRequest defines verify auth challenge request parameters
+type CognitoEventUserPoolsVerifyAuthChallengeRequest struct {
+	UserAttributes             map[string]string `json:"userAttributes"`
+	PrivateChallengeParameters map[string]string `json:"privateChallengeParameters"`
+	ChallengeAnswer            string            `json:"challengeAnswer"`
```
bmoffatt: from https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-verify-auth-challenge-response.html#cognito-user-pools-lambda-trigger-syntax-verify-auth-challenge it looks like this needs to be either an `map[string]string` or an `interface{}`

hgl:
Any chance this getting merge any time soon?

events/cognito.go
mxpv: Fixed

mxpv:
@bmoffatt I'm not quite sure whether `map[string]string` is the right choice since on my test env I was receiving a string type value, so I decided to go with `interface{}`

mxpv:
ping @bmoffatt

bmoffatt:
> @bmoffatt I'm not quite sure whether `map[string]string` is the right choice since on my test env I was receiving a string type value, so I decided to go with `interface{}`

👍

Issue #99, opened 2018-07-26
Support KinesisAnalyticsEvents


defunct73:
Please add support for the KinesisAnalyticsEvents type.

https://github.com/aws/aws-lambda-dotnet/tree/master/Libraries/src/Amazon.Lambda.KinesisAnalyticsEvents

Thanks!

bmoffatt:
merged #169

PR #169, opened 2019-02-18
Feature/add kinesis analytics event

vinujohn:
Added event and response for Kinesis Analytics output delivery.

Followed structure as defined here: https://docs.aws.amazon.com/kinesisanalytics/latest/dev/how-it-works-output-lambda.html

Naming of event and response was copied from .NET and Java: https://docs.aws.amazon.com/kinesisanalytics/latest/dev/how-it-works-output-lambda-functions.html

Added tests.  Refactored existing tests by moving common readJsonFile from dyanamodb_test.go into test package 

Reference issue: #99


bmoffatt:
Thanks!

PR #173, opened 2019-02-27
add AppSyncIdentity

daviskoh:
*Issue #, if available:*

*Description of changes:*

Adds new `AppSyncIdentity` struct which represents the `.identity` field of [context](https://docs.aws.amazon.com/appsync/latest/devguide/resolver-context-reference.html). Other context fields are left out because they are all custom structures that depend on user implementation.

One thing to note is that `AppSyncIdentity.Claims` is of type `map[string]interface{}` which sucks, but Claims values can be either `Number` or `String` json types and there may be custom claims added by the user...


events/appsync.go
```diff
@@ -9,6 +9,20 @@ type AppSyncResolverTemplate struct {
 	Payload   json.RawMessage  `json:"payload"`
 }
 
+// AppSyncIdentity contains information about the caller. The shape of this section depends on the authorization type of your AWS AppSync API
```
bmoffatt: Seems that there'd need to be 2 types added, rather than the union of both shapes. https://docs.aws.amazon.com/appsync/latest/devguide/resolver-context-reference.html
* `AppSyncIAMIdentity`
* `AppSyncCognitoIdentity`

daviskoh: Unmarshalling / marshalling `$context.identity` into the single type works either way, but I'll separate out the type if that's what's desired.

daviskoh:
hmmm also noticing now the test are off...will ping you when resolved.

daviskoh:
@bmoffatt i've updated everything:
- separated out AppSyncIdentity into separate types for IAM & Cognito
- fixed tests to actually check marshalling / unmarshalling of identity json

u mind having a look? 😃

Issue #177, opened 2019-03-17
Links malformed in JSON response


skiarn:
Hello, I have problems returning json payloads containing urls, the encoding/json Marshal will escape & to unicode format. 

Is there a reason why lambda should escape the characters to unicode? 
If not then its better to use json Encoder and turn on SetEscapeHTML(true) or at least make it accessible to configure.

Or am i missing the correct way to format my payload?

My use-case is returning a pre signed s3 url in a json response payload through the lambda but I cant find a way to do that today.

bmoffatt:
Thanks for submitting #178! 

In the meantime, you could implement custom json encoding/decoding for your function by implementing [lambda.Handler](https://github.com/aws/aws-lambda-go/blob/08d251dc58f204411d79012fd6f1beefb7107fea/lambda/handler.go#L14-L16) and passing to [lambda.StartHandler](https://github.com/aws/aws-lambda-go/blob/08d251dc58f204411d79012fd6f1beefb7107fea/lambda/entry.go#L44-L50), which allows receiving and returning raw bytes

``` go
type Handler struct {
  // ...
}
func (h *Handler) Invoke(ctx context.Context, request []byte) ([]byte, error) {
  // ...
}

func main() {
  lambda.StartHandler(Handler)
}
```

bmoffatt:
implemented an option for this in #444

PR #178, opened 2019-03-17
Issue #177 malformed urls inside json response

skiarn:


lambda/handler.go
```diff
@@ -72,6 +94,15 @@ func validateReturns(handler reflect.Type) error {
 	return nil
 }
 
+// SetResponseNonEscapeHTML will NOT coerced JSON strings to valid UTF-8,
+// The angle brackets "<" and ">" will NOT be escaped to "\u003c" and "\u003e"
+// Ampersand "&" is also NOT be escaped to "\u0026".
+func SetResponseNonEscapeHTML(flag bool) {
```
bmoffatt: Instead of a package global setter, I'd prefer optional values to somehow be configured on the type returned by `NewHandler`. I think this could be done in a backwards compatible way in one of these two approaches:

1. adding a `HandlerOptions` type to pass as varargs to `NewHandler`. eg:
  * `NewHandler(handler interface{}, opts... HandlerOptions) lambda.Handler`. 
2. update NewHandler to return a new struct `*JSONHandler` still implementing `lambda.Handler`, and add this add this configuration as a method for that struct. eg:
  * `func NewHandler(handler interface{}) *JSONHandler`
  * `func (handler *JSONHandler) SetEscapeHTML(on bool)` 

I think I like option 2 better, since it would setup a pattern for mirroring the other json encode/decode options like https://golang.org/pkg/encoding/json/#Encoder.SetEscapeHTML

In-use, this'd look like

``` go
func main() {
        jsonHandler := NewHandler(func () (string, error) {
                return "<html><body>html in json string!</body></html>", nil
        });
        jsonHandler.SetEscapeHTML(false);
        lambda.StartHandler(jsonHandler)
}
```

skiarn:
Didnt have to much time lately, took a look at again, do you mean the Handler interface can be modified like ex: https://github.com/skiarn/aws-lambda-go/tree/aws-lambda-encoder
After thinking about it more adding a .SetEscapeHTML(false); method is most likely not very good since users may want to use different encoding/json packages? Therefore I added SetMarshal / SetUnmarshal functions to the Handler interface, but then kind of felt as if using the Invoke function is a better way to go?

Maybe only having `Invoke(ctx context.Context, request []byte) ([]byte, error) ` is good enough?

For me the work around proposed in https://github.com/aws/aws-lambda-go/issues/177, is good enough for me, I have the ability to implement desired serializers. Maybe instead of having more methods on the Handler interface documentation may be better? highlighting how invoke can be used when serializing req and resp?
`func (h *Handler) Invoke(ctx context.Context, request []byte) ([]byte, error) {
  // ...
}

func main() {
  lambda.StartHandler(&Handler{})
}`

PR #185, opened 2019-04-30
Added event for Cognito custom message

JerzyLa:
*Issue #, if available:*

*Description of changes:*
Pull request is based on https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-custom-message.html


events/cognito.go
```diff
@@ -106,3 +106,24 @@ type GroupConfiguration struct {
...
+// allowing a user to customize the message dynamically.
+type CognitoEventUserPoolsCustomMessage struct {
+	CognitoEventUserPoolsHeader
+	Request  CognitoEventUserPoolsCustomMessageRequest  `json:"request"`
+	Response CognitoEventUserPoolsCustomMessageResponse `json:"response"`
+}
+
+// CognitoEventUserPoolsCustomMessageRequest contains the request portion of a CustomMessage event
+type CognitoEventUserPoolsCustomMessageRequest struct {
+	UserAttributes map[string]bool `json:"userAttributes"`
```
bmoffatt: shouldn't this be `map[string]string`? https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-custom-message.html#cognito-user-pools-lambda-trigger-syntax-custom-message

bmoffatt: ugh, I see the sample now - This appears to need to be `map[string]interface{}`

events/cognito.go
```diff
@@ -106,3 +106,24 @@ type GroupConfiguration struct {
...
+	CodeParameter  string          `json:"codeParameter"`
+}
+
+// CognitoEventUserPoolsCustomMessageResponse contains the response portion of a CustomMessage event
+type CognitoEventUserPoolsCustomMessageResponse struct {
+	SmsMessage   string `json:"smsMessage"`
```
bmoffatt: `Sms` -> `SMS`

events/cognito.go
JerzyLa: yes, you're right it should be interface{} instead of bool I will change it. Great thanks for review

events/cognito.go
JerzyLa: Right, good notice

JerzyLa: done

events/cognito.go
JerzyLa: done

PR #187, opened 2019-05-08
Close gzip reader as per documentation

bhavikkumar:
Close the gzip reader as per the [documentation](https://golang.org/pkg/compress/gzip/#NewReader)


bmoffatt:
Thanks!

PR #188, opened 2019-05-09
Add Kinesis Stream Metadata to firehose event

bhavikkumar:
Add sourceKinesisStreamArn and KinesisRecordMetadata to event structs .

Attempting to do a put record API call on a Firehose Delivery stream which has a Kinesis Stream as the source will result in the error `This operation is not permitted on KinesisStreamAsSource delivery stream type.`. Therefore it is useful to have the sourceKinesisStreamArn so put record API calls can be made by Lambda functions to the source Kinesis Stream. 

In addition the Kinesis Record Metadata has been added to the event so the same partition key can be used if the data needs to be reingest to the source Kinesis stream.

This is particularly useful when attempting to reingest data due to hitting the 6MB size limit of the response payload.

### DIFF
--- a/events/firehose.go
+++ b/events/firehose.go
@@ -4,10 +4,11 @@ package events
 
 // KinesisFirehoseEvent represents the input event from Amazon Kinesis Firehose. It is used as the input parameter.
 type KinesisFirehoseEvent struct {
-	InvocationID      string                       `json:"invocationId"`
-	DeliveryStreamArn string                       `json:"deliveryStreamArn"`
-	Region            string                       `json:"region"`
-	Records           []KinesisFirehoseEventRecord `json:"records"`
+	InvocationID           string                       `json:"invocationId"`
+	DeliveryStreamArn      string                       `json:"deliveryStreamArn"`
+	SourceKinesisStreamArn string                       `json:"sourceKinesisStreamArn"`
+	Region                 string                       `json:"region"`
+	Records                []KinesisFirehoseEventRecord `json:"records"`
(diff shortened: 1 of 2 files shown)


events/firehose.go
```diff
@@ -32,3 +34,10 @@ type KinesisFirehoseResponseRecord struct {
 	Result   string `json:"result"` // The status of the transformation. May be TransformedStateOk, TransformedStateDropped or TransformedStateProcessingFailed
 	Data     []byte `json:"data"`
 }
+
+type KinesisFirehoseRecordMetadata struct {
+	ShardID                     string                `json:"shardId"`
+	PartitionKey                string                `json:"partitionKey"`
+	SequenceNumber              string                `json:"sequenceNumber"`
+	ApproximateArrivalTimestamp MilliSecondsEpochTime `json:"approximateArrivalTimestamp"`
```
bmoffatt: The CI failed (https://travis-ci.org/aws/aws-lambda-go/builds/530382884). Need to check if the type needs to be `string`, or if the test data is wrong.
```
...
--- FAIL: TestSampleTransformation (0.00s)
    firehose_test.go:50: could not unmarshal event. details: json: cannot unmarshal string into Go struct field KinesisFirehoseRecordMetadata.approximateArrivalTimestamp of type int64
...
```

bmoffatt: found a reference internally and a sample event in the console - I will update the test data and merge

bmoffatt:
Thanks!

PR #189, opened 2019-05-12
Add go module support

### LINKED ISSUE #181: Go modules support
Could you please add `go.mod` and `go.sum` files in the repository root? Thanks!

bhavikkumar:
This resolves #181 

Initialized go modules from go deps. Modified the testRuntimeStackTrace test to compare slice of the package path instead of the raw string. as only the last 5 packages/directories are returned from frame.path.

Modified the Travis-CI job to have a job which sets GO111MODULE to on and overrides the before_install and install step to use `go mod download` instead of `dep ensure`. All other steps are kept the same. See https://travis-ci.com/bhavikkumar/aws-lambda-go/builds/111472734

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,8 +3,9 @@ module github.com/bhavikkumar/aws-lambda-go
 go 1.12
 
 require (
-	github.com/davecgh/go-spew v1.1.0
-	github.com/pmezard/go-difflib v1.0.0
+	github.com/aws/aws-lambda-go v1.10.0
+	github.com/davecgh/go-spew v1.1.0 // indirect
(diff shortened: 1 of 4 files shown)


philtay:
The first line of `go.mod` should be `github.com/aws/aws-lambda-go`, not `github.com/bhavikkumar/aws-lambda-go`.

bhavikkumar:
Good spot. I have fixed the module name.

go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/aws/aws-lambda-go
```
bmoffatt: LGTM! thanks!

PR #192, opened 2019-05-16
Enable function to be created outside of StartHandler.

efritz:
Function is useless outside of the lambda package as there is no way to create a function with a non-nil handler. I have need to, for example, change the error conditions from log.Fatal within the entrypoint, but not having access to a Function constructor stops me from writing a reasonable replacement in my application code.

### DIFF
--- a/lambda/entry.go
+++ b/lambda/entry.go
@@ -53,9 +53,7 @@ func StartHandler(handler Handler) {
 	if err != nil {
 		log.Fatal(err)
 	}
-	function := new(Function)
-	function.handler = handler
-	err = rpc.Register(function)
+	err = rpc.Register(NewFunction(handler))
 	if err != nil {
 		log.Fatal("failed to register handler function")
 	}
--- a/lambda/function.go
+++ b/lambda/function.go
@@ -16,6 +16,10 @@ type Function struct {
 	handler Handler
 }
 
+func NewFunction(handler Handler) *Function {
+	return &Function{handler: handler}
+}
+
 func (fn *Function) Ping(req *messages.PingRequest, response *messages.PingResponse) error {
 	*response = messages.PingResponse{}
 	return nil


bmoffatt:
>  I have need to, for example, change the error conditions from log.Fatal within the entry-point

Super curious about your use-case! I suppose your plan also includes and handle `rpc.Register` and `rpc.Accept` yourself?

efritz:
> Super curious about your use-case! I suppose your plan also includes calling `rpc.Register` and `rpc.Accept` yourself?

It's not all that interesting - we're using a custom logger for use with localstack on our development machines and the messages from log.Fatal don't come through in a way that's immediately inspectable for us. The version we have right now simply returns an error instead of killing the process, and that error is handled by our application entrypoint.

Although now that we've done this, we see that sometimes the bind call is trying to re-use the same port within the lambda container from a previous invocation of the function, so we may end up adding a small backoff to net.Listen.

bmoffatt:
>  we see that sometimes the bind call is trying to re-use the same port within the lambda container from a previous invocation of the function

eg: logs like this?
> 2019/03/19 18:56:09 listen tcp 127.0.0.1:53604: bind: address already in use
> 2019/03/19 18:56:09 exit status 1 

Are you *currently* seeing this behavior? We recently deployed a change in the service to mitigate a similar sounding issue we saw in our own internal tests.

> It's not all that interesting - we're using a custom logger for use with localstack on our development machines and the messages from log.Fatal don't come through in a way that's immediately inspectable for us.

Gotcha. Out of curiosity, is this an OSS logging formatter like https://github.com/sirupsen/logrus, or is it something else? Is this something that could be solved by using https://golang.org/pkg/log/#SetOutput?

efritz:
Yes, those logs. It turns out that it was bad behavior on the part of localstack plus the image lambci/lambda:go1.x. It's an issue between the workflow between these two parts. Localstack issues a docker exec command into binary wrapping the lambda handler in the lambda:go1.x image. This image will start a new process for each invocation and kill it once the lambda function completes. This makes the container start a new listener on the same hard-coded port for each invocation, where they clash under any amount of concurrency.

Rewriting the wrapper in that container solves our issue (and we're trying to push that fix upstream). I don't think this issue has anything to do with this SDK (although the visibility made possible by this change was nice).

> Out of curiosity, is this an OSS logging formatter like https://github.com/sirupsen/logrus, or is it something else?

It's using [gomol](https://github.com/aphistic/gomol).

> Is this something that could be solved by using https://golang.org/pkg/log/#SetOutput?

There are lots of solutions, but we were using a TCP/JSON output logger to stream logs in real time via socat (and could then grep through it via jq). We come form an environment where we were using an ELK stack when our stuff was hosted out-of-cloud.

bmoffatt:
Ah, guess I had missed the emphasis on the "local" bit. Thanks for bearing with the questions, we're always interested in the details of how our customers are developing applications.

Issue #196, opened 2019-05-24
Lex Sample Bug: lex.go does not define LexResponse


scconroy:
The sample here: https://github.com/aws/aws-lambda-go/blob/master/events/README_Lex.md

References:

*lex.LexResponse

However, this struct is not defined in lex.go:

https://github.com/aws/aws-lambda-go/blob/master/events/lex.go

Creating a custom struct worked as a workaround:

type LexResponse struct {
SessionAttributes map[string]string `json:"sessionAttributes,omitempty"` 
DialogAction *events.LexDialogAction `json:"dialogAction,omitempty"` 
}

PR #197, opened 2019-05-24
Add lex response

ezhdan-nyt:
Issue #196 

*Add missed LexResponse struct, update events README*


bmoffatt:
matches https://docs.aws.amazon.com/lex/latest/dg/lambda-input-response-format.html#using-lambda-response-format

Thanks for the fix!

PR #208, opened 2019-06-26
Add Lambda handler wrapper for SNS-based custom resources

roberth-k:
*Description of changes:*
Adds `cfn.LambdaWrapSNS()`. This function is similar to `cfn.LambdaWrap()`, but it first unwraps the CloudFormation custom resource input payload from a SNS event. An invalid SNS event is considered an invocation error.


bmoffatt:
Can you point me to an example (docs, other languages, etc..) of how the SNS gets sent to the Lambda? 

1. It's not clear to me how/why this would get used instead of the direct Lambda backed customer resources.
2. I want to do a functional test myself :)

roberth-k:
[This guide](https://aws.amazon.com/blogs/mt/multi-account-strategy-using-aws-cloudformation-custom-resources-to-create-amazon-route-53-resources-in-another-account/) from the AWS Blog provides a practical example, although some of the samples are erroneous (e.g. the Message of an example event not being shown as a string). [Amazon's documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources-sns.html) for SNS-backed custom resources does suggest the input signature is the same as for Lambda-backed custom resources, but having recently implemented one I've had to include this SNS event wrapper to make it work.

I suspect that a SNS-backed custom resource is invoked via the Publish API rather than any special integration, so it would make sense that the event is delivered with the appropriate decorations.

bmoffatt:
That blog really helps put this in context, thanks!
> You might be asking yourself why I used SNS when CloudFormation custom resources can invoke Lambda directly. I could have invoked Lambda directly, but SNS simplifies cross account permissions and makes the configuration easier.

PR #211, opened 2019-07-02
Extend events.SimpleEmailReceiptAction fields

griffithsh:
In my team, we regularly use SES events, but we Unmarshal into a custom type because the version in this repository omits fields that we require. The fields we use are documented here: https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-notifications-contents.html#receiving-email-notifications-contents-action-object

Specifically the `bucketName` and `objectKey` fields on the `action` object, which as documented, are only present when the type of the action is `S3`. In this PR, I've included _all_ documented fields in the SimpleEmailReceiptAction type. This will allow my team to delete our custom type and use a type from an official source when Unmarshaling SES events.


griffithsh:
I'm thinking there should be more extensive tests included as part of this PR, as the structure of the type is significantly more complex. Is a change like this likely to be accepted?

bmoffatt:
> I'm thinking there should be more extensive tests included as part of this PR, as the structure of the type is significantly more complex. Is a change like this likely to be accepted?

A summary like as a godoc comment would help, as would adding a few more tests cases to [ses_test.go](https://github.com/aws/aws-lambda-go/blob/master/events/ses_test.go)

[testdata/ses-event.json](https://github.com/aws/aws-lambda-go/blob/master/events/testdata/ses-event.json) contains a `Lambda` type event, I think SNS and S3 types will be covered by copying [these samples from the AWS docs](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-notifications-examples.html).

Let me know if you'd like to do these updates, otherwise I'll try to do them soonish.

griffithsh:
That's great, I can spend some more time now working through your feedback.

griffithsh:
I think these three new commits should address your suggestions.

bmoffatt:
Thanks!

PR #212, opened 2019-07-02
Update S3 Event to working Go file

AaronMulgrew:
*Issue #, if available:*
Example provided does not work without changing the script slightly...

*Description of changes:*
Added main function and corrected imports to make it a working Go file (that can be compiled/built)


bmoffatt:
Thanks!

PR #183, opened 2019-04-13
Add fields to CognitoEventUserPoolsPostConfirmationResponse

booomch:
Added fields to response model (CognitoEventUserPoolsPostConfirmationResponse) for custom messages, that field exists in serialized json below but did not exists in struct CognitoEventUserPoolsPostConfirmationResponse.

{
    "version": "1",
    "triggerSource": "CustomMessage_ForgotPassword",
    "region": "us-east-1",
    "userPoolId": "XXXX",
    "callerContext": {
        "awsSdkVersion": "aws-sdk-unknown-unknown",
        "clientId": "XXXX"
    },
    "userName": "XXX",
    "request": {
        "userAttributes": {
            "cognito:email_alias": "XXX@XXX.com",
            "cognito:user_status": "CONFIRMED",
            "email": "XXX@XXX.com",
            "email_verified": "true",
            "name": "XXX@XXX.com",
            "preferred_username": "XXX",
            "sub": "XXXXX"
        }
    },
    "response": {
        "emailMessage": nil,
        "emailSubject": nil,
        "smsMessage": nil
    }
}

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #217, opened 2019-08-01
[add] Add fields to CognitoEventUserPoolsPostConfirmationResponse

ken880guchi:
### Sorry. please read it first:
This pull request is equivalent to #183 but an exam has been added.
This change is something that I want to be included as soon as possible.

If there is no problem, I would like you to accept this pull request. 🙇 

### Issue #, if available:
Can not change the subject, message and sms message sent in a custom response.

### Description of changes:
Add a field to `CognitoEventUserPoolsPostConfirmationResponse` class
This is useful for changing the text and subject and sms message send in a custom response.

### DIFF
--- a/events/cognito.go
+++ b/events/cognito.go
@@ -96,6 +96,9 @@ type CognitoEventUserPoolsPostConfirmationRequest struct {
 
 // CognitoEventUserPoolsPostConfirmationResponse contains the response portion of a PostConfirmation event
 type CognitoEventUserPoolsPostConfirmationResponse struct {
+	EmailMessage string `json:"emailMessage"`
+	EmailSubject string `json:"emailSubject"`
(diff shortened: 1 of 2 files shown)


bmoffatt:
Do you have a link handy to an example or documentation? I found only the [user-pool-custom-message]( https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-custom-message.html) documentation, which appears different to the [post-confirmation-response](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-post-confirmation.html) documentation, which doesn't indicate response fields.

ken880guchi:
@bmoffatt Thank you for your reply.

From your tips, I was misunderstanding

Close this pull request.

PR #224, opened 2019-08-15
Adds OperationName to API Gateway Proxy Request

jmnarloch:
operationName is included by API Gateway for API's that were imported through OpenAPI spec.

*Issue #, if available:*
N/A
*Description of changes:*
API Gateway APIGatewayProxyRequestContext varies depending on the way how the API is being created. If I use `aws apigateway import-rest-api` [See](https://docs.aws.amazon.com/cli/latest/reference/apigateway/import-rest-api.html). The model of the event used to call Lambda will contain additional attribute.

One of such attributes is `operationName` that will be populated from OpenAPI operationId attribute.

This pull request adds this additional field to allow handling it on the Lambda side.


jmnarloch:
@jackrk I will rebase the change and update the pull request.

Issue #186, opened 2019-05-07
DynamoDB attributevalue.Integer() loses precision for large values


BastiPaeltz:
When dealing with large integer values (timestamps in nanoseconds) that you are able to store in DynamoDB, `attributevalue.Integer()` will lose precision when converting to int64. An example value would be `1556893286484340854`.  
This is caused by the `strconv.ParseFloat(value, 64)` LOC that this function uses before converting to int64. When doing a `strconv.ParseInt(value, 10, 64)` instead, the value does not lose precision.

diegobernardes:
This code can be used to confirm this behaviour:

```golang
package main

import (
	"fmt"
	"strconv"
	"time"
)

func main() {
	base := time.Now().UnixNano()
	baseString := strconv.FormatInt(base, 10)

	floatParse, err := strconv.ParseFloat(baseString, 64)
	if err != nil {
		panic(err)
	}
	floatResult := int64(floatParse)

	intResult, err := strconv.ParseInt(baseString, 10, 64)
	if err != nil {
		panic(err)
	}

	fmt.Println("origin:             ", base)
	fmt.Println("current behaviour:  ", floatResult)
	fmt.Println("proposed behaviour: ", intResult)
}
```

Execution result:
```
origin:              1570726625494837000
current behaviour:   1570726625494836992
proposed behaviour:  1570726625494837000
```

diegobernardes:
Solved at https://github.com/aws/aws-lambda-go/pull/227.

PR #227, opened 2019-08-29
fixed loses precision for large values

joeke80215:
*Issue #, if available:*
#186
*Description of changes:*
loses precision for large values


joeke80215:
ok, thanks your suggest

Issue #228, opened 2019-09-03
CodeBuildEvent struct not matching CloudWatchEvents


whithajess:
**CloudWatchEvents** seem to not match the structure of **CodeBuildEvents** and when performing `json.Unmarshal()` throws an error:
```
json: cannot unmarshal number 60.0 into Go struct field CodeBuildEventAdditionalInformation.timeout-in-minutes of type int64: UnmarshalTypeError
null
```
It looks like Test Data (with minor changes) was pulled from: [AWS Build Notification Docs](https://docs.aws.amazon.com/codebuild/latest/userguide/sample-build-notifications.html#sample-build-notifications-ref) which may have been correct at the time but what I am observing being returned from CloudWatchEvents is `60.0` not `60`

You should be able to recreate with a simple lambda listening to CloudWatchEvents with source: `"aws.codebuild"`
```
package main

import (
	"github.com/aws/aws-lambda-go/events"
	"github.com/aws/aws-lambda-go/lambda"
)

func CodeBuildNotifier(event events.CodeBuildEvent) {}

func main() {
	lambda.Start(CodeBuildNotifier)
}
```

This looks like it can be fixed easily by changing the types in https://github.com/aws/aws-lambda-go/blob/master/events/duration.go to `float64` instead of `int64` PR to follow.

whithajess:
@roberth-k I noticed you only recently put this in so you may have objections or already run up against this - any feedback would be appreciated.

roberth-k:
@whithajess I did come across the same issue, but only fixed it locally (https://github.com/roberth-k/aws-lambda-go/commit/8f0289add20a55cb2e49439e81ff40f230ce1dbe) and did not put up a PR at the time. The PR looks good to me.

PR #229, opened 2019-09-03
Fix Precision issue with Durations

whithajess:
* CloudWatchEvents seem to return with precision 60.0 instead of 60
* Change tests to match
* Change types to float64

#228 


whithajess:
👍 made changes to have both in tests.

whithajess:
@bmoffatt You may want to merge this one as you put through the original change set. Thanks!

Issue #244, opened 2019-10-28
CodePipelineEvent hardcoded to CodePipelineJob.


whithajess:
**Is your feature request related to a problem? Please describe.**

`CodePipelineEvent` seems to be hardcoded as a `CodePipelineJob`
- https://github.com/aws/aws-lambda-go/blob/master/events/codepipeline_job.go#L6-L8

This incorrectly makes CodePipeline event unusable for its actual events.

**Describe the solution you'd like**

I would like to move the current `CodePipelineEvent` into its own `CodePipelineJobEvent` as it seems to be seperate then the documented CodePipelineEvents

**Describe alternatives you've considered**

Considered just leaving it as is and creating `CodePipelineEvent` under some new name like `ActualCodePipelineEvent` but I think this is one of those instances where it would be better to just make the change.

**Additional context**
**Breaking Change**

whithajess:
@bmoffatt Just raising this again.

After 4 years this issue still exists, and anyone who uses this `aws-lambda-go` and `CodePipelineEvent` actually gets `CodePipelineJobEvent` completely unexpected and not matching the actual event or the rest of this codebase.
  
Is holding back doing SemVer changes really worth this? Wouldn't it have just been better to bump it?

I'm not currently doing anything with the library so it doesn't matter to me, but figured it might be worth reflecting on.

PR #245, opened 2019-10-28
[BREAKING CHANGE] Clean up CodePipeline Job Implementation

whithajess:
* Incorrectly used "CodePipelineEvent"

*Issue #, if available:*

#244

*Description of changes:*

Moves `CodePipelineJob` into its own `CodePipelineJobEvent` giving back the `CodePipelineEvent` namespace for actual usage.


whithajess:
@bmoffatt Did you have any thoughts on this? Thanks.

bmoffatt:
> @bmoffatt Did you have any thoughts on this? Thanks.

I'd rather not do a semver-major change. I'm ok with adding and alias `type CodePipelineEvent = CodePipelineJobEvent` to help with the ambiguity.

whithajess:
@bmoffatt I understand the reluctance for a `semver-major` but this will trip a lot of people up in the future even with the alias, it breaks the principle of least surprise all the other Events are what they say they are. This is what I did originally - Assumed that `CodePipelineEvent` was the same as `CodeBuildEvent` and then when digging in found that `CodePipelineEvent` was something completely unexpected!

bmoffatt:
We won't be merging intentionally breaking changes at this time. I've tagged #244 as requiring a v2 release to keep track of the request.

For #247, a different type name can be used.

PR #266, opened 2020-02-29
Update build-lambda-zip to use urfave/cli/v2

alex:
*Description of changes:*

Updates build-lambda-zip to use the v2 of urfave/cli. This is the newer, maintained, version of that package.


bmoffatt:
Thanks!

Issue #236, opened 2019-09-18
_X_AMZN_TRACE_ID is not set to the correct value


hamitb:
Other language runtimes in Lambda provide X-Ray Trace information through the `_X_AMZN_TRACE_ID` environment variable. Recently this trace information also started to be appended to the Cloudwatch logs for each of the invocations as in the below:

![image](https://user-images.githubusercontent.com/16473976/65144154-5536b200-da1f-11e9-91fa-b60b56168b8e.png)

Even though in the other runtimes(Python, Node.js, etc.) the trace information provided through the environment variable and through the Cloudwatch logs seems to be consistent, this is not the case for the Go.

Following lambda function can be used to reproduce the issue: 
```go
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/aws/aws-lambda-go/lambda"
)

type myEvent struct {
	ID string `json:"id"`
}

func handler(ctx context.Context, event myEvent) (string, error) {
	fmt.Println(os.Getenv("_X_AMZN_TRACE_ID"))
	fmt.Println(ctx.Value("x-amzn-trace-id"))

	return "Hello ƛ!", nil
}

func main() {
	lambda.Start(handler)
}
```

When I run the above lambda function, I got the values below:
![image](https://user-images.githubusercontent.com/16473976/65145152-857f5000-da21-11e9-9366-3e9395885696.png)

Here the values in the rectangle 2 and 3 are consistent as expected. However, the value provided by the environment variable(rectangle 1) does not seem to be correct.

hamitb:
`x-amzn-trace-id` is put into context here. Maybe `_X_AMZN_TRACE_ID` environment variable can also be set to the same value here if it's an appropriate place to do so.
https://github.com/aws/aws-lambda-go/blob/02a2e049b01224f55250fc6738badc312874b250/lambda/function.go#L61

bmoffatt:
released with v1.15.0

PR #269, opened 2020-03-05
Update _X_AMZN_TRACE_ID on every invoke

bmoffatt:
*Issue #, if available:*

#236 

*Description of changes:*

Updates the value of the environment variable _X_AMZN_TRACE_ID on every invoke. This is consistent with [the docs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-x-ray.html#lambda-x-ray-env-variables)

### DIFF
--- a/lambda/function.go
+++ b/lambda/function.go
@@ -5,6 +5,7 @@ package lambda
 import (
 	"context"
 	"encoding/json"
+	"os"
 	"reflect"
 	"time"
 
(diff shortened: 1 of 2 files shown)


alex:
I don't think this will work. The Go Lambda support uses net/rpc to server traffic. net/rpc spawns a go-routine per incoming connection.

Unless there's no concurrency at all here, because only a single connection is ever established, these environment variables will become a race condition, constantly stomping over each other between concurrent connections. Environemnt variables in general cannot be used for state for this reason.

If it really is the case that there's no concurrency, I'd implore you to add a comment next to the `Setenv` call.

mhart:
A single Lambda (unix) process only ever processes one request at a time

PR #279, opened 2020-04-03
for -o to work, you specify it as an alias rather than in the Name

marnoldinfinity:
*Description of changes:*
Move 'o' to an alias when declaring the flag


bmoffatt:
Thanks for the fix!

PR #280, opened 2020-04-07
Add APIGatewayV2HTTPResponse

stefansundin:
*Issue #, if available:*

*Description of changes:*
- ~~There is a new version field for payload format version 1.0. The value is `"1.0"` as expected.~~ Removed this commit since this field only exists in HTTP API (it is present in both payload format version 1.0 and 2.0).
- In payload format version 2.0, you can send cookies in an easier way with the new cookies field. This does not work in 1.0, so I think we want a new struct for that, right?

I used this page as a reference: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html

Maybe someone more knowledgable than me should double check that there aren't other fields that are missing?

### DIFF
--- a/events/apigw.go
+++ b/events/apigw.go
@@ -93,6 +93,16 @@ type APIGatewayV2HTTPRequestContextHTTPDescription struct {
 	UserAgent string `json:"userAgent"`
 }
 
+// APIGatewayV2HTTPResponse configures the response to be returned by API Gateway V2 for the request
+type APIGatewayV2HTTPResponse struct {
+	StatusCode        int                 `json:"statusCode"`
+	Headers           map[string]string   `json:"headers"`
+	MultiValueHeaders map[string][]string `json:"multiValueHeaders"`
+	Body              string              `json:"body"`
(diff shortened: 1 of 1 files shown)


stefansundin:
I did my tests above with my HTTP API. Now when the tests failed, I realized that I also have to test it with the old REST API. I did so and it appears that the Version field is _not_ set there. So I think I should remove that commit and maybe we should be happy with just adding the new response struct?

Let me know what you think.

ewbankkit:
@stefansundin Looking at the docs - https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format - it seems like the API Gateway v1 (REST APIs) Lambda integration event structure didn't change, so I think NOT adding `Version` to `APIGatewayProxyRequest` is the correct decision.

ewbankkit:
@stefansundin Can you add a test response json file and corresponding test case? Maybe this will help with review.

bmoffatt:
Thanks for the contribution!

PR #283, opened 2020-04-16
use best zip compression

jftuga:
This patch updates build-lambda-zip such that the max compression level is used instead of the default zip compression level.


bmoffatt:
Note: I added a test in #307 that can log how this change affects the .zip file

bmoffatt:
I accidentally hit "close" instead of "merge" - but I see that the source branch no longer exists. Feel free to re-submit this PR

PR #287, opened 2020-04-27
Support setting an optional base context for functions.

krak3n:
*Issue #, if available:* 

n/a

---
*Description of changes:* 

Adds support for setting a base context to be given to functions that take context as their first argument. Current implementation is to provide a `context.Background()`. This implements a similar feature as seen in `net/http` where a `http.Server` can be given a `BaseContext` function which returns a `context.Context` to be given to all http requests: https://golang.org/src/net/http/server.go#L2572

This avoids the need for injecting context values into the context before a function is called.

This is a non breaking API change as new explicit methods have been added to support passing in a base context.

---

### DIFF
--- a/lambda/function.go
+++ b/lambda/function.go
@@ -16,29 +16,12 @@ import (
 // Function struct which wrap the Handler
 type Function struct {
 	handler Handler
-	context context.Context
+	ctx     context.Context
 }
 
 // NewFunction which creates a Function with a given Handler
 func NewFunction(handler Handler) *Function {
 	return &Function{handler: handler}
-}
-
-// NewFunctionWithContext which creates a Function with a given Handler and sets the base Context.
-func NewFunctionWithContext(ctx context.Context, handler Handler) *Function {
-	return &Function{
-		context: ctx,
-		handler: handler,
-	}
-}
-
-// Context returns the base context used for the fn.
-func (fn *Function) Context() context.Context {
-	if fn.context == nil {
-		return context.Background()
-	}
-
-	return fn.context
 }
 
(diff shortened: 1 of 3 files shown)


lambda/function.go
```diff
@@ -16,13 +16,31 @@ import (
...
+	context context.Context
 }
 
 // NewFunction which creates a Function with a given Handler
 func NewFunction(handler Handler) *Function {
 	return &Function{handler: handler}
 }
 
+// NewFunctionWithContext which creates a Function with a given Handler and sets the base Context.
+func NewFunctionWithContext(ctx context.Context, handler Handler) *Function {
```
bmoffatt, line 28: Can this be private? `NewFunction` doesn't really have a reason to be in the public API, so it'd be nice to not increase surface area further.

lambda/function.go
```diff
@@ -16,13 +16,31 @@ import (
...
+	return &Function{
+		context: ctx,
+		handler: handler,
+	}
+}
+
+// Context returns the base context used for the fn.
+func (fn *Function) Context() context.Context {
```
bmoffatt, line 36: This should be private, or moved inside of `Invoke`. The public visibility of the other `Function` methods is done only as a requirement of the `net/ipc` library.

lambda/function.go
krak3n: @bmoffatt removed this function. I decided to go for the `http.Request.WithContext` approach instead, except that `WithContext` is not exported on the `Function`. I think that's a little cleaner :+1:

lambda/function.go
krak3n: @bmoffatt made this un exported :+1:

PR #288, opened 2020-04-27
Add SNSMessageAttribute type

hardbap:
Implements `SNSMessageAttribute` which was inspired by the [SQSMessageAttribute](https://github.com/aws/aws-lambda-go/blob/master/events/sqs.go#L22).

Since this is a breaking change I'm not expecting it to get merged but I figured I'd submit it anyhow.

### DIFF
--- a/events/sns.go
+++ b/events/sns.go
@@ -18,15 +18,20 @@ type SNSEventRecord struct {
 }
 
 type SNSEntity struct {
-	Signature         string                 `json:"Signature"`
-	MessageID         string                 `json:"MessageId"`
-	Type              string                 `json:"Type"`
-	TopicArn          string                 `json:"TopicArn"`
-	MessageAttributes map[string]interface{} `json:"MessageAttributes"`
-	SignatureVersion  string                 `json:"SignatureVersion"`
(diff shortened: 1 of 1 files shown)


events/sns.go
```diff
@@ -18,15 +18,20 @@ type SNSEventRecord struct {
...
+	MessageAttributes map[string]SNSMessageAttribute `json:"MessageAttributes"`
+	SignatureVersion  string                         `json:"SignatureVersion"`
+	Timestamp         time.Time                      `json:"Timestamp"`
+	SigningCertURL    string                         `json:"SigningCertUrl"`
+	Message           string                         `json:"Message"`
+	UnsubscribeURL    string                         `json:"UnsubscribeUrl"`
+	Subject           string                         `json:"Subject"`
+}
+
+type SNSMessageAttribute struct {
```
bmoffatt, line 34: Type is missing `Name` per the docs: https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html

bmoffatt: oops, misunderstood, the key is the name :)

bmoffatt:
I like the change, but you're right, I don't believe it can be merged without doing a V2 release

Issue #294, opened 2020-06-02
Failing to parse APIGateway V2 HTTP API payload


LaurenceGA:
Suddenly all of my projects using this library broke :disappointed: 
I have HTTP APIs all using payload version 2.0 and the two req/resp structs from this library (APIGatewayV2HTTPRequest/APIGatewayProxyResponse). They also have JWT authorisers attached.

They all crash immediately with a JSON parsing error:
`json: cannot unmarshal bool into Go struct field APIGatewayV2HTTPRequestContextAuthorizerJWTDescription.requestContext.authorizer.jwt.claims of type string: UnmarshalTypeError
null`

I had to convert all of my projects back to HTTP API payload format version 1.0 to fix this issue, so I think it's something with specifically parsing v2.0 payload.

I guess breaking changes were made to the payload somehow that this library can't handle?

BrentSouza:
I have a similar issue. 

* HTTP API
* Route with JWT Authorizer

It was working previously, but now I get the following error:

```
json: cannot unmarshal number into Go struct field .requestContext.authorizer.jwt.claims of type string: UnmarshalTypeError null
```

BrentSouza:
So I think it's just that claims needs to be a `map[string]interface{}` since claims like `exp` and `iat` are numbers. Not sure how this worked before unless the API Gateway code was rewriting them all as strings...

bmoffatt:
I've passed this issue along to the service team. You may also want to open a case with AWS support.

LaurenceGA:
> I've passed this issue along to the service team. You may also want to open a case with AWS support.

Thanks!
Looks like everything is magically back to working actually.
I guess the service team reverted the change or something 🤷 

I guess that will make #295 no longer required 🤔 

If so, I guess it's fine to close this issue and the PR. Might need some confirmation from the service team about what happened and if any such changes might occur in the future that would effect this library?

LaurenceGA:
I think I found the culprit, got an email from AWS support some time ago:
> Starting May 4, 2020, there will be a change how we present JWT claims fields in the Lambda context variable [1].
Currently, all primitive types are automatically converted to String objects when passed into the JWT claims field. For example, if you had passed “sampleKey”:1, the context variable would have “sampleKey”:”1”.
We will update this behavior to write the primitive types as they are passed in and no longer automatically convert them to String objects. For example, if you had passed “sampleKey”:1, the context variable would have “sampleKey”:1.

[...]

PR #295, opened 2020-06-03
Allow non-string value claims in APIGatewayV2HTTPRequestContextAuthorizerJWTDescription

BrentSouza:
*Issue #, if available:*

#294 

*Description of changes:*

For some reason events from an HTTP API with a JWT Authorizer can no longer unmarshal claims.  Not sure if something upstream in API Gateway changed, but making this a map[string]interface{} allows number values like the iat or exp claims to unmarshal without error. 


bmoffatt:
Closing. Will leave #294 while I confirm things with the service team.

PR #302, opened 2020-06-28
feat: add s3 batch result code constants

moofish32:
This adds constants and a type for S3 Batch Result Codes similar to what
is available for CodeBuild results

### DIFF
--- a/events/s3_batch_job.go
+++ b/events/s3_batch_job.go
@@ -2,6 +2,16 @@
 
 package events
 
+// S3BatchJobResultCode represents the result of an S3BatchJobTask (i.e.
+// succeeded, permanent failure, or temmporary failure)
+type S3BatchJobResultCode string
+
+const (
+	S3BatchJobResultCodeSucceeded        S3BatchJobResultCode = "Succeeded"
+	S3BatchJobResultCodeTemporaryFailure                      = "TemporaryFailure"
+	S3BatchJobResultCodePermanentFailure                      = "PermanentFailure"
+)
+
 // S3BatchJobEvent encapsulates the detail of a s3 batch job
 type S3BatchJobEvent struct {
 	InvocationSchemaVersion string           `json:"invocationSchemaVersion"`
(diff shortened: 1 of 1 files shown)


events/s3_batch_job.go
```diff
@@ -33,7 +45,7 @@ type S3BatchJobResponse struct {
 
 // S3BatchJobResult represents the result of a given task
 type S3BatchJobResult struct {
-	TaskID       string `json:"taskId"`
-	ResultCode   string `json:"resultCode"`
-	ResultString string `json:"resultString"`
+	TaskID       string               `json:"taskId"`
+	ResultCode   S3BatchJobResultCode `json:"resultCode"`
```
bmoffatt, line 49: I think adding the type creates a breaking change. 

ex: https://play.golang.org/p/T5G1eKtbBor

I think leaving the type as `string` is fine

bmoffatt (changes requested):

events/s3_batch_job.go
```diff
@@ -2,6 +2,18 @@
 
 package events
 
+// S3BatchJobResultCode represents the result of an S3BatchJobTask (i.e.
+// succeeded, permanent failure, or temmporary failure)
+type S3BatchJobResultCode string
```
bmoffatt, line 7: change to a type alias `type S3BatchJobResultCode = string`

events/s3_batch_job.go
```diff
@@ -2,6 +2,18 @@
...
+
+// S3BatchJobResultCode represents the result of an S3BatchJobTask (i.e.
+// succeeded, permanent failure, or temmporary failure)
+const (
+	S3BatchJobResultCodeSucceeded        S3BatchJobResultCode = "Succeeded"
```
bmoffatt, line 12: downside of changing `S3BatchJobResult.ResultCode` type back to `string`, is that comparisons against this new type require a cast!
```
if inputEvent.Results[0].ResultCode == S3BatchJobResultCodeSucceeded {
  // ..
}
```
-> 
```
Invalid operation: inputEvent.Results[0].ResultCode == S3BatchJobResultCodeSucceeded (mismatched types string and S3BatchJobResultCode)
```

Instead, go back to having S3BatchJobResult being a `S3BatchJobResultCode`. It is useful for self documentation purposes. But also change the type to be a type alias eg: `type S3BatchJobResultCoce = string` - this will preserve backwards compatibility.

bmoffatt (changes requested):

PR #307, opened 2020-07-10
cmd/build-lambda-zip: Produce valid .zips for provided runtime functions

bmoffatt:
*Description of changes:*

This is a follow-up to #298, which added support for using Go with the `provided` runtime. This change updates the build-lambda-zip tool.

In addition to making the first argument executable, a symlink named "bootstrap" will also be placed in the root of the .zip file, and will point to the executable.

doc reference: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html

### DIFF
--- a/cmd/build-lambda-zip/main.go
+++ b/cmd/build-lambda-zip/main.go
@@ -1,3 +1,5 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved
+
 package main
 
 import (
@@ -5,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 
@@ -37,6 +40,7 @@ func main() {
 			if err := compressExeAndArgs(outputZip, inputExe, c.Args().Tail()); err != nil {
 				return fmt.Errorf("failed to compress file: %v", err)
 			}
+			log.Print("wrote " + outputZip)
 			return nil
 		},
 	}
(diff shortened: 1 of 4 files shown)


cmd/build-lambda-zip/main.go
```diff
@@ -48,6 +52,19 @@ func main() {
 }
 
 func writeExe(writer *zip.Writer, pathInZip string, data []byte) error {
+
```
carlzogh, line 55: nit: extra new line

cmd/build-lambda-zip/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSizes(t *testing.T) {
+	if testing.Short() {
+		t.Skip()
+		return
+	}
+	t.Log("Test how different arguments affect binary and archive sizes")
```
carlzogh, line 24: nit: lowercase `Test` for log format consistency

cmd/build-lambda-zip/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+		require.NoError(t, err)
+		defer zipReader.Close()
+	eachFile:
+		for _, path := range filePaths[1:] {
+			for _, zipFileEntry := range zipReader.File {
+				if zipFileEntry.Name == path {
+					continue eachFile
+				}
+			}
+			t.Logf("Failed to find resource file %s in zip", path)
```
carlzogh, line 136: nit: lowercase `Failed` for log format consistency

bmoffatt: done

cmd/build-lambda-zip/main_test.go
bmoffatt: done

cmd/build-lambda-zip/main.go
bmoffatt: done

PR #312, opened 2020-08-11
lambda: check if error type is InvokeResponse_Error

### LINKED ISSUE #308: detect if returned error is of type InvokeResponse_Error
We currently check the `errorType` field to determine the category of errors to which a message corresponds. However, the `errorType` field is set by reflecting on the error value's concrete type, which leads to an influx of `errorString` or other error types that show up. Since, Go doesn't support creating a named type (https://github.com/golang/go/issues/16522), we need some way of setting a custom `errorType` value.

The lambda handler already constructs an `InvokeResponse_Error`, after reflecting on the user-returned error. Just type-assert that the error's underlying concrete type is `InvokeResponse_Error`, and if so, use that. This will allow us to directly set its `Type` field, as necessary.

Currently, we work around this by predeclaring a whole host of "dummy" error wrapper types, like so:

```Go
type categoryA struct{ error }
func (e categoryA) Error() string   { return e.error.Error() }
type categoryB struct{ error }
func (e categoryB) Error() string   { return e.error.Error() }
type categoryC struct{ error }
func (e categoryC) Error() string   { return e.error.Error() }
type categoryD struct{ error }
func (e categoryD) Error() string   { return e.error.Error() }
...
```

smasher164:
Currently, we cannot set the errorType field without creating a custom
named error type. This causes a proliferation of dummy named error types
whose only purpose is to set a text field, and limits the errorType
field to what can be expressed as an identifier.

This change checks that the panicked or returned error is if type
InvokeResponse_Error, and if so, uses it directly in the invoker's
response. InvokeResponse_Error is now updated to implement error.

Note: This is fully backwards compatible, since InvokeResponse_Error
previously did not implement the error interface.

Fixes #308.

### DIFF
--- a/lambda/errors.go
+++ b/lambda/errors.go
@@ -17,6 +17,9 @@ func getErrorType(err interface{}) string {
 }
 
 func lambdaErrorResponse(invokeError error) *messages.InvokeResponse_Error {
+	if ive, ok := invokeError.(messages.InvokeResponse_Error); ok {
+		return &ive
+	}
 	var errorName string
 	if errorType := reflect.TypeOf(invokeError); errorType.Kind() == reflect.Ptr {
 		errorName = errorType.Elem().Name()
@@ -30,6 +33,9 @@ func lambdaErrorResponse(invokeError error) *messages.InvokeResponse_Error {
 }
 
 func lambdaPanicResponse(err interface{}) *messages.InvokeResponse_Error {
+	if ive, ok := err.(messages.InvokeResponse_Error); ok {
+		return &ive
+	}
 	panicInfo := getPanicInfo(err)
 	return &messages.InvokeResponse_Error{
 		Message:    panicInfo.Message,
--- a/lambda/handler_test.go
+++ b/lambda/handler_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	"github.com/aws/aws-lambda-go/lambda/handlertrace"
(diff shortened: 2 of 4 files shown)


bmoffatt:
Thanks for the PR @smasher164 - Until now, I've considered the types in the `messages` package to be an internal implementations detail, but that's not a strongly held opinion. 

I've forwarded this to the team to check for alternative approaches to the problem.

smasher164:
Hey @bmoffatt, friendly ping to ask if there's any word on the issue/PR?
I have a couple of ideas on alternative approaches:
- The exported type does not have to be from the `messages` package. You can introduce and detect another named type. For instance, you could introduce something like this in the `lambda` package:
```Go
type Error struct{
    Message string `json:"errorMessage"`		
    Type    string `json:"errorType"`
}
```
- Introduce an interface that the user's desired error type could satisfy. This would be an optional interface that would require a type assertion, like:
```Go
type Error interface {
    error
    Type() string
}
```
The disadvantage is that the second option is a potentially breaking change, if the user's error implements the Type method.

bmoffatt:
I talked with the team, and we figured that if we ever want to get rid of the `messages` library, we'd have to do a `v2` release regardless of your initial proposal. I'll merge the change as-is

The `lambda.Error` change you proposed is cosmetically nicer, but I don't like that we'd just be introducing more types just to end up copying into InvokeResponse_Error anyways.

bmoffatt (approved):

PR #313, opened 2020-08-13
Add missing request ID to context in norpc mode

MattWhelan:
When self-hosting a Go lambda, the Request ID was missing from the invocation context.


bmoffatt:
Thanks for catching this! I'm adding a test case to this PR before merging

PR #325, opened 2020-09-19
reuse runtimeAPIClient.buffer

shogo82148:
*Issue #, if available:*

`runtimeAPIClient.buffer` is defined on here https://github.com/aws/aws-lambda-go/blob/55dc88be4cdfaaf4bb4d9f65debd8d4310e0a83c/lambda/runtime_api_client.go#L32
but it is never used.

*Description of changes:*

`runtimeAPIClient.next()` is never called concurrently, so we can reuse `runtimeAPIClient.buffer`.
It avoids extra re-allocation.


bmoffatt:
Nice catch, thanks! This was the original intent, but I think I missed it during an internal refactor :)

bmoffatt (approved):

PR #326, opened 2020-09-19
rename safeMarshal to mustMarshal

shogo82148:
*Issue #, if available:*

the `safeMarshal` function might return invalid JSON if `err.Error()` contains special characters of JSON.

https://github.com/aws/aws-lambda-go/blob/55dc88be4cdfaaf4bb4d9f65debd8d4310e0a83c/lambda/invoke_loop.go#L103-L109

*Description of changes:*

use `json.Marshal` instead of `fmt.Sprintf`.

### DIFF
--- a/lambda/invoke_loop.go
+++ b/lambda/invoke_loop.go
@@ -99,10 +99,13 @@ func convertInvokeRequest(invoke *invoke) (*messages.InvokeRequest, error) {
 	return res, nil
 }
 
+// mustMarshal is like json.Marshal but panics if the v cannot be marshaled.
+// use this if you know the v doesn't contain values that json.Marshal can't marshal,
+// such as Channel, complex, and function values.
 func mustMarshal(v interface{}) []byte {
 	payload, err := json.Marshal(v)
 	if err != nil {
-		panic(err)
+		panic("lambda: " + err.Error())
 	}
 	return payload
 }
--- a/lambda/invoke_loop.go
+++ b/lambda/invoke_loop.go
@@ -47,7 +47,7 @@ func handleInvoke(invoke *invoke, function *Function) error {
 	}
 
 	if functionResponse.Error != nil {
-		payload := safeMarshal(functionResponse.Error)
+		payload := mustMarshal(functionResponse.Error)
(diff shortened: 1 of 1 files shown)


bmoffatt:
can you include a test case that exercises this?

lambda/invoke_loop.go
```diff
@@ -100,10 +99,13 @@ func convertInvokeRequest(invoke *invoke) (*messages.InvokeRequest, error) {
...
 
-func safeMarshal(v interface{}) []byte {
+// mustMarshal is like json.Marshal but panics if the v cannot be marshaled.
+// use this if you know the v doesn't contain values that json.Marshal can't marshal,
+// such as Channel, complex, and function values.
+func mustMarshal(v interface{}) []byte {
 	payload, err := json.Marshal(v)
 	if err != nil {
-		return []byte(fmt.Sprintf(serializationErrorFormat, err.Error()))
+		panic("lambda: " + err.Error())
```
shogo82148, line 108: Sorry, I've misunderstood.
`json.Marshal(v)` doesn't return any error, because we know `*messages.InvokeResponse_Error` can be marshaled.
So, we don't need to treat this case, and just enough to panic.

shogo82148: "Must"-prefix is usually used in Go (e.g. https://golang.org/pkg/regexp/#MustCompile https://golang.org/pkg/html/template/#Must )

bmoffatt:
I don't think `panic` is the right thing to do here

* `"Runtime.SerializationError"` is no longer reported to the data-plane, so the client of the function doesn't know what went wrong
* `panic`ing also causes the next invoke to be a a kind of cold start, which doesn't seem appropriate if the only problem was a response serialization error

shogo82148:
I tried to write an AWS Lambda handler that causes `"Runtime.SerializationError"` for testing, but I couldn’t.
There is no way to cause `"Runtime.SerializationError"`, so calling `panic` is simplest way to handle the error.

> the `safeMarshal` function might return invalid JSON if `err.Error()` contains special characters of JSON.

This problem remains unresolved.
I'll try another fix.

PR #327, opened 2020-09-24
safeMarshal should escape the error message

shogo82148:
*Issue #, if available:*

the `safeMarshal` function might return invalid JSON if `err.Error()` contains special characters of JSON. ref https://github.com/aws/aws-lambda-go/pull/326

https://github.com/aws/aws-lambda-go/blob/55dc88be4cdfaaf4bb4d9f65debd8d4310e0a83c/lambda/invoke_loop.go#L103-L109

*Description of changes:*

use `json.Marshal` instead of `fmt.Sprintf`.

### DIFF
--- a/lambda/invoke_loop_test.go
+++ b/lambda/invoke_loop_test.go
@@ -106,6 +106,18 @@ func TestReadPayload(t *testing.T) {
 
 }
 
+type invalidPayload struct{}
+
+func (invalidPayload) MarshalJSON() ([]byte, error) {
+	return nil, errors.New(`some error that contains '"'`)
+}
+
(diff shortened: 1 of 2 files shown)


bmoffatt (approved):

lambda/invoke_loop_test.go
```diff
@@ -106,6 +106,18 @@ func TestReadPayload(t *testing.T) {
 
 }
 
+type invalidPayload struct{}
+
+func (invalidPayload) MarshalJSON() ([]byte, error) {
```
bmoffatt, line 111: ahhh! thanks for this test, it explains the problem well :)

Issue #82, opened 2018-05-25
Problem With S3 Event


toneill818:
The S3Object struct has URLDecodedKey but it does not seem to be populated when processing request. Looking at the [example](https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html) event message structure this is not returned. The Key field is populated but spaces in filenames are replace with a '+'. Is this the expected behavior?

oyi812:
Shoddy. Key should be the unencoded value. There should be an implemented field (looking at you AWS lambda devs) called URLEncodedKey. I'm certain this has caused problems for a fair number of users. It should have been fixed long ago.

ARolek:
I just hit this problem as well. Unfortunate bug.

estahn:
I'm facing the same issue.

mkozjak:
I just hit the same issue. Might be related to Object ACL settings?

vsomayaji:
The URL-decoded value is not included in the event. The SDK should be setting `URLDecodedKey` similar to how the Java SDK does it:

https://github.com/aws/aws-sdk-java/blob/6a4c873c71320ef0175ca1c13188e9c850a85e51/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/event/S3EventNotification.java#L176-L183

Until it does, you could set it yourself using something like this:

```go
	key, _ := url.QueryUnescape(record.S3.Object.Key)
	record.S3.Object.URLDecodedKey = key
```

PR #335, opened 2020-11-11
Remove URLDecodedKey from S3Object

johanneswuerbach:
*Issue #, if available:* Fixes https://github.com/aws/aws-lambda-go/issues/82

*Description of changes:* Populates the `URLDecodedKey` property from the `S3Object` struct as it is currently empty and is never populated by S3 itself https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html

The population implements the convenience method as the lambda Java SDK https://github.com/aws/aws-sdk-java/blob/6a4c873c71320ef0175ca1c13188e9c850a85e51/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/event/S3EventNotification.java#L176-L183

### DIFF
--- a/events/s3.go
+++ b/events/s3.go
@@ -3,6 +3,8 @@
 package events
 
 import (
+	"encoding/json"
+	"net/url"
 	"time"
 )
 
@@ -52,6 +54,20 @@ type S3Object struct {
 	VersionID     string `json:"versionId"`
 	ETag          string `json:"eTag"`
 	Sequencer     string `json:"sequencer"`
+}
+
+func (o *S3Object) UnmarshalJSON(data []byte) error {
+	type rawS3Object S3Object
+	if err := json.Unmarshal(data, (*rawS3Object)(o)); err != nil {
+		return err
+	}
+	key, err := url.QueryUnescape(o.Key)
+	if err != nil {
+		return err
+	}
+	o.URLDecodedKey = key
+
+	return nil
 }
 
 type S3TestEvent struct {
--- a/events/s3_test.go
+++ b/events/s3_test.go
@@ -26,8 +26,11 @@ func TestS3EventMarshaling(t *testing.T) {
 		t.Errorf("could not marshal event. details: %v", err)
 	}
 
-	// 4. check result
-	assert.JSONEq(t, string(inputJSON), string(outputJSON))
+	// 4. read expected output JSON from file
+	exepectedOutputJSON := test.ReadJSONFromFile(t, "./testdata/s3-event-with-decoded.json")
+
+	// 5. check result
+	assert.JSONEq(t, string(exepectedOutputJSON), string(outputJSON))
 }
 
 func TestS3TestEventMarshaling(t *testing.T) {
--- a/events/testdata/s3-event-with-decoded.json
+++ b/events/testdata/s3-event-with-decoded.json
@@ -28,6 +28,7 @@
         },
         "object": {
           "key": "Happy%20Face.jpg",
+          "urlDecodedKey": "Happy Face.jpg",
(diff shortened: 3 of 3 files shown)


bmoffatt:
Instead of removing it, maybe we can fix it with a custom deserializer

johanneswuerbach:
Happy to do that, but shouldn’t it be a method computed demand then or should it be computed every time even if not needed?

For backwards compatibility the later seems better, but feels kind of wasteful. Preferences?

harrisonhjones:
We could benchmark it but my guess is that the performance impact of keeping it would be very small. I would personally prefer for the package to keep to the Go backwards compatibility promise as much as is possible / reasonable.

johanneswuerbach:
Updated to always populate the property instead of removing it.

bmoffatt:
I think that we should also change the json tag, to help signal to other that urlDecodedKey is not a part of the event model
```
type S3Object {
        ....
	URLDecodedKey string `json:"-"` // populated by custom deserializer
        ....
}
```

johanneswuerbach:
I was also thinking about that, but discarded the idea as that also seems like a breaking change. If someone serialises the current object to json, this would modify the shape of the json.

harrisonhjones:
> I was also thinking about that, but discarded the idea as that also seems like a breaking change. If someone serialises the current object to json, this would modify the shape of the json.

Unfortunate but probably true.

I wonder if we should consider keeping a list of "stuff we should consider fixing in a MV 2"? Maybe create an issue with some kind of "mv2" tag?

bmoffatt:
We can't guarantee full compatibility for consumers of the serialized version of the object, arguably it breaks either way.
1. if the tag stays, consumer sees `{"urlDecodedKey": "" }` -> `{"urlDecodedKey": "some value"}`
2. if the field is ignored, consumer sees `{"urlDecodedKey": ""}` -> `{}`

CC @carlzogh what do you think? Have we run into this situation in the java project recently? My preference right now, is to take breaking changes if the type if we would otherwise be documenting something false. Were we code-generating these, and the schema updated, I think we'd default into that behavior.

> > I was also thinking about that, but discarded the idea as that also seems like a breaking change. If someone serialises the current object to json, this would modify the shape of the json.
> 
> Unfortunate but probably true.
> 
> I wonder if we should consider keeping a list of "stuff we should consider fixing in a MV 2"? Maybe create an issue with some kind of "mv2" tag?

@harrisonhjones I've been using the `requires-v2` tag for this

carlzogh:
The change here brings this library to parity with the Java events library (ref. [`urlDecodedKey`](https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/models/s3/S3EventNotification.java#L123-L151)) and I think this is the right approach.

From a consumer's perspective, I don't see this as a very risky change as a property that was previously empty now holds a value. We're not changing the JSON contract - were we to remove the field, I'd see that as a breaking change and would probably think twice about it. 

I can't think of consumer use-cases that would break as a result of us merging this in, aside from the fact that we're not differentiating between service event models and fields that `aws-lambda-go` adds in to these events (eg. `urlDecodedKey`). This being a pre-existing issue that is not made worse by this PR, I wouldn't hold up merging this in but I still think it's worth trying to figure out how we can better draw this line - potentially in v2.

bmoffatt:
Agree on not holding up this change, since it doesn't make the existing problem worse :) 

I'll move my thoughts on https://github.com/aws/aws-lambda-go/pull/335#issuecomment-736015096 to a separate issue to track

bmoffatt (approved):

PR #343, opened 2020-12-11
Add a ClientVPN connection handler request/response definition

gpoul:
Also added a sample connection handler and linked it from the README

### DIFF
--- a/events/README_ClientVPN.md
+++ b/events/README_ClientVPN.md
@@ -26,13 +26,13 @@ func handler(request events.ClientVPNConnectionHandlerRequest) (events.ClientVPN
 
 	sourceIP := request.PublicIP
 	if net.ParseIP(sourceIP) == nil {
-		return events.ClientVPNConnectionHandlerResponse{}, fmt.Errorf("Invalid parameter PublicIP passed in request: '%s'", sourceIP)
+		return events.ClientVPNConnectionHandlerResponse{}, fmt.Errorf("Invalid parameter PublicIP passed in request: %q", sourceIP)
 	}
 
-	log.Printf("SOURCE IP: %s", sourceIP)
+	log.Printf("SOURCE IP: %q", sourceIP)
 
 	if allowed, ok := AllowedIPs[sourceIP]; ok && allowed {
-		log.Printf("Allowing access from: %s", sourceIP)
+		log.Printf("Allowing access from: %q", sourceIP)
 		return events.ClientVPNConnectionHandlerResponse{
 			Allow: true,
 			ErrorMsgOnFailedPostureCompliance: "",
@@ -41,10 +41,10 @@ func handler(request events.ClientVPNConnectionHandlerRequest) (events.ClientVPN
 		}, nil
 	}
 
-	log.Printf("Blocking access from: %s", sourceIP)
+	log.Printf("Blocking access from: %q", sourceIP)
 	return events.ClientVPNConnectionHandlerResponse{
 		Allow: false,
-		ErrorMsgOnFailedPostureCompliance: "You're accessing from a blocked IP range.",
+		ErrorMsgOnFailedPostureCompliance: "You're trying to connect from an IP address that is not allowed.",
 		PostureComplianceStatuses: []string{"BlockedSourceIP"},
 		SchemaVersion: "v1",
 	}, nil
--- a/events/README_ClientVPN.md
+++ b/events/README_ClientVPN.md
@@ -1,0 +1,59 @@
+# Sample Function
+
+The following is a sample Lambda function that receives a Client VPN connection handler request as an input and then validates the IP address input and checks whether the connection source IP is on the allowed list defined as a map inside the function. If the source IP matches an allowed IP address it allows the access, otherwise an error message is presented to the user. Debug logs are generated to CloudWatch Logs. (Note that by default anything written to Console will be logged as CloudWatch Logs events.)
+
+```go
+import (
+	"fmt"
+	"log"
+	"net"
+
+	"encoding/json"
+
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/aws/aws-lambda-go/lambda"
+)
+
+var (
+	AllowedIPs = map[string]bool{
+		"10.11.12.13": true,
+	}
+)
+
+func handler(request events.ClientVPNConnectionHandlerRequest) (events.ClientVPNConnectionHandlerResponse, error) {
+	requestJson, _ := json.MarshalIndent(request, "", "  ")
+	log.Printf("REQUEST: %s", requestJson)
+
+	sourceIP := request.PublicIP
+	if net.ParseIP(sourceIP) == nil {
+		return events.ClientVPNConnectionHandlerResponse{}, fmt.Errorf("Invalid parameter PublicIP passed in request: '%s'", sourceIP)
+	}
+
+	log.Printf("SOURCE IP: %s", sourceIP)
+
+	postureCompliance := []string{}
+
+	allowed, ok := AllowedIPs[sourceIP]
+	if !ok {
+		allowed = false
+		postureCompliance = []string{"BlockedSourceIP"}
+	}
+
+	if allowed {
+		log.Printf("Allowing access from: %s", sourceIP)
+	} else {
+		log.Printf("Blocking access from: %s", sourceIP)
+	}
+
+	return events.ClientVPNConnectionHandlerResponse{
+		Allow: allowed,
+		ErrorMsgOnFailedPostureCompliance: "You're accessing from a blocked IP range.",
+		PostureComplianceStatuses: postureCompliance,
+		SchemaVersion: "v1",
+	}, nil
+}
+
+func main() {
+	lambda.Start(handler)
+}
+```
--- a/events/README_ClientVPN.md
+++ b/events/README_ClientVPN.md
@@ -31,24 +31,21 @@ func handler(request events.ClientVPNConnectionHandlerRequest) (events.ClientVPN
 
 	log.Printf("SOURCE IP: %s", sourceIP)
 
-	postureCompliance := []string{}
-
-	allowed, ok := AllowedIPs[sourceIP]
-	if !ok {
-		allowed = false
-		postureCompliance = []string{"BlockedSourceIP"}
-	}
-
-	if allowed {
+	if allowed, ok := AllowedIPs[sourceIP]; ok && allowed {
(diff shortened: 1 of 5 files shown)


events/README_ClientVPN.md
```diff
@@ -0,0 +1,59 @@
...
+	"net"
+
+	"encoding/json"
+
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/aws/aws-lambda-go/lambda"
+)
+
+var (
+    AllowedIPs = map[string]bool{
```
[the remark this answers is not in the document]
gpoul: fixed this one; good catch!

events/README_ClientVPN.md
```diff
@@ -0,0 +1,59 @@
+# Sample Function
+
+The following is a sample Lambda function that receives a Client VPN connection handler request as an input and then validates the IP address input and checks whether the connection source IP is on the allowed list defined as a map inside the function. If the source IP matches an allowed IP address it allows the access, otherwise an error message is presented to the user. Debug logs are generated to CloudWatch Logs. (Note that by default anything written to Console will be logged as CloudWatch Logs events.)
```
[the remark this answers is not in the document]
gpoul: This is mentioned in 17 other README_*.md files; I agree it's technically incorrect. Would you agree to fix all of those to:

`(Note that anything written to stdout or stderr will be logged as CloudWatch Logs events.)`

events/README_ClientVPN.md
```diff
@@ -0,0 +1,59 @@
...
+	if net.ParseIP(sourceIP) == nil {
+		return events.ClientVPNConnectionHandlerResponse{}, fmt.Errorf("Invalid parameter PublicIP passed in request: '%s'", sourceIP)
+	}
+
+	log.Printf("SOURCE IP: %s", sourceIP)
+
+	postureCompliance := []string{}
+
+	allowed, ok := AllowedIPs[sourceIP]
+	if !ok {
```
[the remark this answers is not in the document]
gpoul: The reason I didn't want to do this is that I wanted a single place to create the response, but I can see how this is cleaner with multiple returns; but I'd be more comfortable with having the path within the if condition be the allowed one and returning a failure in all other cases. I'll get back to you with a suggestion and re-test this as well.

events/README_ClientVPN.md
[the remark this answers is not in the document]
harrisonhjones: Seems reasonable to me.

events/README_ClientVPN.md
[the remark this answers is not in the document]
gpoul: @harrisonhjones I've now made this change in a1e5abc

events/README_ClientVPN.md
```diff
@@ -0,0 +1,59 @@
...
+
+	if allowed {
+		log.Printf("Allowing access from: %s", sourceIP)
+	} else {
+		log.Printf("Blocking access from: %s", sourceIP)
+	}
+
+	return events.ClientVPNConnectionHandlerResponse{
+		Allow: allowed,
+		ErrorMsgOnFailedPostureCompliance: "You're accessing from a blocked IP range.",
```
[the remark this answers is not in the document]
gpoul: I'm not sure; the "You're accessing from a blocked IP range" is more likely to resonate with the user's point of view as he can successfully connect from another IP address.

"Your IP address is not allowed" sounds more like we authenticate only based on the IP address.

If I switch this to allow it would be something like "You're not accessing from an allowed IP range" which I think is more confusing than the original with the double-negation. What do you think?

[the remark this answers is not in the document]
harrisonhjones: I'm not going to block on it so whatever you want is fine. From my point of view this is an example, we are likely not going to change it later to expand the list of allowed ips or to add more complicated denial logic. It currently allows/denies based on the client's IP only so messaging that reflects that makes sense to me.

events/README_ClientVPN.md
```diff
@@ -0,0 +1,56 @@
...
+		log.Printf("Allowing access from: %s", sourceIP)
+		return events.ClientVPNConnectionHandlerResponse{
+			Allow: true,
+			ErrorMsgOnFailedPostureCompliance: "",
+			PostureComplianceStatuses: []string{},
+			SchemaVersion: "v1",
+		}, nil
+	}
+
+	log.Printf("Blocking access from: %s", sourceIP)
```
harrisonhjones, line 44: nit pick: thoughts on using `%q` instead of `%s` so that the IP here (and in other lines?) is automatically quoted?

gpoul: fixed; I'm sorry but I didn't even know %q was a thing. Thanks for taking the time to review this and point it out!

events/README_ClientVPN.md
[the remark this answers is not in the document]
gpoul: I've adjusted the message as well to make it clearer with what we're doing; I hope.

harrisonhjones (approved):
LGTM. @bmoffatt what do you think?

events/README_ClientVPN.md
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	log.Printf("SOURCE IP: %q", sourceIP)
+
+	if allowed, ok := AllowedIPs[sourceIP]; ok && allowed {
+		log.Printf("Allowing access from: %q", sourceIP)
+		return events.ClientVPNConnectionHandlerResponse{
+			Allow: true,
+			ErrorMsgOnFailedPostureCompliance: "",
+			PostureComplianceStatuses: []string{},
```
harrisonhjones, line 39: small nitpick: do you need to set either of these? The nil value for a string is "" and it might be ok for `PostureComplianceStatuses` to be nil?

gpoul: I've written a test like this to see how I can make this happen:

```go
func TestClientVPNConnectionHandlerResponseConstructorValidation(t *testing.T) {
        // read json from file
        inputJSON, err := ioutil.ReadFile("./testdata/clientvpn-connectionhandler-response.json")
        if err != nil {
                t.Errorf("could not open test file. details: %v", err)
        }

        // construct response
        outputEvent := ClientVPNConnectionHandlerResponse{
                Allow: true,
                SchemaVersion: "v1",
        }

        // serialize to json
        outputJSON, err := json.Marshal(outputEvent)
        if err != nil {
                t.Errorf("could not marshal event. details: %v", err)
        }

        assert.JSONEq(t, string(inputJSON), string(outputJSON))
}
```

for the string value that is easy; I can just drop it and it will be `""`, but the array is not so easy and I need it in there; if I understand https://github.com/golang/go/issues/37711 correctly the choice is either a `func NewClientVPNConnectionHandlerResponse` or a custom marshaller that handles this; both feel like they depart from how the rest of the library is used a bit, although there seems to be functions for DynamoDB attributes as well.

Anything you'd suggest? I somehow feel it's easier to keep this in the sample and hopefully people will use the sample to then end up with the right data structure? The alternative that feels clean as well is the function that creates an "empty" object as it could also create the SchemaVersion as well and we can drop the version from the sample code; that would gain something at least.

What do you think?

harrisonhjones:
```go
 ClientVPNConnectionHandlerResponse{
                Allow: true,
                SchemaVersion: "v1",
        }
```

should marshal as:

```json
{
	"allow": true,
	"error-msg-on-failed-posture-compliance": "",
	"posture-compliance-statuses": null,
        "schema-version": "v1",
}
```

The question is: is this ok? Or does it cause failures. Does it need to be:

```json
{
	"allow": true,
	"error-msg-on-failed-posture-compliance": "",
	"posture-compliance-statuses": [],
        "schema-version": "v1",
}
```

If it does then yea, we'll need to keep the:

```go
PostureComplianceStatuses: []string{},
```

events/README_SQS.md
```diff
@@ -1,7 +1,7 @@
 
 # Sample Function
 
-The following is a sample class and Lambda function that receives Amazon SQS event message data as input, writes some of the message data to CloudWatch Logs, and responds with a 200 status and the same body as the request. (Note that by default anything written to Console will be logged as CloudWatch Logs events.)
```
bmoffatt, line 4: Appreciate the docs changes. But they aren't related to the ClientVPN feature. Can you split them into a separate PR?

events/codepipeline_test.go
```diff
@@ -18,7 +18,7 @@ func TestCodePipeLineEventMarshaling(t *testing.T) {
 		t.Errorf("could not open test file. details: %v", err)
 	}
 
-	// de-serialize into CognitoEvent
```
bmoffatt, line 21: Thanks for noticing this, but it's best if the PR we merge is focused only on the new feature. Feel free to open a separate PR that bundles up the docs and comment fixes you've found.

events/README_ClientVPN.md
gpoul: It tested with

```go
                return events.ClientVPNConnectionHandlerResponse{
                        Allow: true,
                        ErrorMsgOnFailedPostureCompliance: "",
                        SchemaVersion: "v1",
                }, nil
```

and

```go
                return events.ClientVPNConnectionHandlerResponse{
                        Allow: true,
                        SchemaVersion: "v1",
                }, nil
```

and they both fail with an "internal" error. I think the `{}` is indeed required to be returned to ClientVPN.

Would you think a `func NewClientVPNConnectionHandlerResponse` would be good or should we leave the example as-is?

events/README_SQS.md
gpoul: I have cherry-picked the other changes to a separate branch and will shortly push those changes

events/codepipeline_test.go
gpoul: I have cherry-picked the other changes to a separate branch and will shortly push those changes

events/README_ClientVPN.md
harrisonhjones: Ah, that stinks. It's `[]` not `{}` right?

gpoul: Yes, sorry; it's an empty array. Even in JSON. The behavior with ClientVPN I've tested is that both code snippets above in https://github.com/aws/aws-lambda-go/pull/343#discussion_r545870909 don't result in a successful authentication; whereas if I use the original snippet that initializes an empty array instead of nil it works.

So I think we either need to include this in the sample or provide a function that correctly initializes the response struct.

bmoffatt (approved):

bmoffatt:
thanks for the contribution!

PR #346, opened 2020-12-17
Add Amazon MQ event structure

shask-amazon:
*Issue #, if available:*

*Description of changes:*
Adds event structure for Amazon Active MQ.

*Testing
Unit tested.
Created a go lambda function with ActiveMQEvent as input event and validated event deserialization by creating an event source mapping for an ActiveMQ queue on Amazon ActiveMQ broker.

### DIFF
--- a/events/activemq.go
+++ b/events/activemq.go
@@ -3,28 +3,28 @@
 package events
 
 type ActiveMQEvent struct {
-	EventSource    string              `json:"eventSource"`
-	EventSourceArn string              `json:"eventSourceArn"`
-	Messages       []ActiveMQMessage   `json:"messages"`
+	EventSource    string            `json:"eventSource"`
+	EventSourceArn string            `json:"eventSourceArn"`
+	Messages       []ActiveMQMessage `json:"messages"`
 }
 
 type ActiveMQMessage struct {
-	MessageID       string          `json:"messageID"`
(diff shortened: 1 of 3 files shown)


bmoffatt:
updated some naming conventions that aren't caught by the CI workflows

events/activemq.go
```diff
@@ -0,0 +1,30 @@
...
+	Timestamp     int64               `json:"timestamp"`
+	DeliveryMode  int                 `json:"deliveryMode"`
+	CorrelationID string              `json:"correlationID"`
+	ReplyTo       string              `json:"replyTo"`
+	Destination   ActiveMQDestination `json:"destination"`
+	Redelivered   bool                `json:"redelivered"`
+	Type          string              `json:"type"`
+	Expiration    int64               `json:"expiration"`
+	Priority      int                 `json:"priority"`
+	Data          string              `json:"data"`
```
bmoffatt, line 23: @shask-amazon from the docs and sample data, this seems to always be base64 encoded? What do you think about changing the type to `[]byte` so that the encode/decode is automatic for users?

shask-amazon: Fair point. I chose string to be consistent with POJO -  https://tinyurl.com/y8wde5pn

bmoffatt (approved):

bmoffatt (approved):

PR #350, opened 2021-01-09
Add `omitempty` to `HTTPMethod` in `APIGatewayProxyRequest`

LegNeato:
This can be empty for websocket events, see https://github.com/LegNeato/aws-lambda-events/issues/33

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/events/apigw.go
+++ b/events/apigw.go
@@ -6,7 +6,7 @@ package events
 type APIGatewayProxyRequest struct {
 	Resource                        string                        `json:"resource"` // The resource path defined in API Gateway
 	Path                            string                        `json:"path"`     // The url path for the caller
-	HTTPMethod                      string                        `json:"httpMethod"`
+	HTTPMethod                      string                        `json:"httpMethod,omitempty"`
 	Headers                         map[string]string             `json:"headers"`
 	MultiValueHeaders               map[string][]string           `json:"multiValueHeaders"`
 	QueryStringParameters           map[string]string             `json:"queryStringParameters"`


harrisonhjones:
I'm just curious: why do you want this? Are you marshalling this struct to JSON? Why?

LegNeato:
I generate rust code from the go code: https://github.com/LegNeato/aws-lambda-events/issues/33

bmoffatt:
did you mean to modify [APIGatewayWebsocketProxyRequest](https://github.com/aws/aws-lambda-go/blob/b3dd246b7d83fd62856549646711e5328075c69a/events/apigw.go#L149) instead? 

Since it's unused, I think we could deprecate the field, and remove the json tag. Would that play well with your code generator?

LegNeato:
I'm pretty sure I modified the correct struct. If you look at the corresponding C# class (https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayProxyRequest.cs) they point out that different fields can be null for this event depending on if the source is http or websockets.

I just look at the structs, so if there is additional marshalling logic elsewhere that chooses between the two structs please let me know and I'll see what I can do on my side.

bmoffatt:
There's no marshaling logic, it's up the the author of the function to select the correct struct for the event source. The [java events project also has a separate type for web socket events](https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2WebSocketEvent.java)

calavera:
@LegNeato it might make more sense to modify the Websocket structure like @bmoffatt suggests. Leaving up to the user is kind of unfortunate because for them both requests come from origin. It will probably be reasonable to supplement with specific examples of usage for each one of the structs.

bmoffatt:
superseded by #355

PR #352, opened 2021-01-23
Use http.Header for MultiValueHeaders

RaeesBhatti:
*Issue #, if available:*

*Description of changes:*

Use `http.Header` type for `MultiValueHeaders` field in the following types:
* `APIGatewayProxyRequest`
* `APIGatewayProxyResponse`
* `APIGatewayV2HTTPResponse`
* `APIGatewayWebsocketProxyRequest`
* `APIGatewayCustomAuthorizerRequestTypeRequest`

The change is backwards compatible and makes it easier to work with headers in requests and responses. By allowing developers to do `response.MultiValueHeaders.Set("Content-Type", "plain/text")` or better yet `response.MultiValueHeaders.Add("Set-Cookie", "ping=pong")`

### DIFF
--- a/events/apigw.go
+++ b/events/apigw.go
@@ -24,11 +24,11 @@ type APIGatewayProxyRequest struct {
 
 // APIGatewayProxyResponse configures the response to be returned by API Gateway for the request
 type APIGatewayProxyResponse struct {
-	StatusCode        int                 `json:"statusCode"`
-	Headers           map[string]string   `json:"headers"`
-	MultiValueHeaders http.Header         `json:"multiValueHeaders"`
-	Body              string              `json:"body"`
-	IsBase64Encoded   bool                `json:"isBase64Encoded,omitempty"`
+	StatusCode        int               `json:"statusCode"`
+	Headers           map[string]string `json:"headers"`
+	MultiValueHeaders http.Header       `json:"multiValueHeaders"`
+	Body              string            `json:"body"`
+	IsBase64Encoded   bool              `json:"isBase64Encoded,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


bmoffatt:
while useful, this change would not be backwards compatible.

``` go
package main

import (
	"fmt"
	"net/http"
)

type Response struct {
	headers http.Header // program fails to compile
	// headers map[string][]string // program compiles
}

type MyMapType map[string][]string

func mutate(x MyMapType) {
	x["hello"] = []string{"hello"}
}

var _ = http.StatusOK // prevent unused import error

func main() {
	r := &Response{headers: make(map[string][]string)}
	mutate(r.headers)
	fmt.Printf("r = %v\n", r)
}
```
```
./prog.go:23:10: cannot use r.headers (type http.Header) as type MyMapType in argument to mutate

Go build failed.
```
ref: https://play.golang.org/p/Wh_3UIjeYFH


It's a little bit more typing, but you can still get the UX you desire in your code by adding a type cast
``` go
package main

import (
	"fmt"
	"net/http"
)

type Response struct {
	MultiValueHeaders map[string][]string
}

func main() {
	r := &Response{MultiValueHeaders: make(map[string][]string)}
	http.Header(r.MultiValueHeaders).Add("Set-Cookie", "ping=pong")
	http.Header(r.MultiValueHeaders).Add("Set-Cookie", "foo=bar")
	fmt.Printf("r = %v\n", r)
}
```
```
r = &{map[Set-Cookie:[ping=pong foo=bar]]}
```
ref: https://play.golang.org/p/Bfd_-ZO5Z2K

PR #364, opened 2021-03-03
Update tests.yaml to use build matrix

furusax0621:
GitHub Actions supports using a build matrix, it can run tests in multiple versions.

I suggest this implementation because using this method is easier to maintain.


bmoffatt:
Nice, thanks!

PR #373, opened 2021-04-22
Allow non-string value in body response for api gateway

ytobing:
*Issue #, if available:* -

*Description of changes:*

change data type for `body` in `APIGatewayProxyResponse` and `APIGatewayV2HTTPResponse` struct to interface so it will be more flexible to return the data, instead of only able to return it as a string. 

### DIFF
--- a/events/apigw.go
+++ b/events/apigw.go
@@ -23,7 +23,7 @@ type APIGatewayProxyResponse struct {
 	StatusCode        int                 `json:"statusCode"`
 	Headers           map[string]string   `json:"headers"`
 	MultiValueHeaders map[string][]string `json:"multiValueHeaders"`
-	Body              string              `json:"body"`
+	Body              interface{}         `json:"body"`
 	IsBase64Encoded   bool                `json:"isBase64Encoded,omitempty"`
 }
 
@@ -123,7 +123,7 @@ type APIGatewayV2HTTPResponse struct {
 	StatusCode        int                 `json:"statusCode"`
 	Headers           map[string]string   `json:"headers"`
 	MultiValueHeaders map[string][]string `json:"multiValueHeaders"`
-	Body              string              `json:"body"`
+	Body              interface{}         `json:"body"`
 	IsBase64Encoded   bool                `json:"isBase64Encoded,omitempty"`
 	Cookies           []string            `json:"cookies"`
 }


MrGossett:
This is already possible with the `IsBase64Encoded` attribute. Maybe you're looking for a helper func that will marshal an `interface{}` to a strong value so it can be set in the body? But what that helper would need to know if you're marshaling as JSON, XML, ASN.1, or just encoding raw bytes as Base64.

ytobing:
@MrGossett not like that actually. In my case i want to put a struct in that body, instead of put JSON string to that body

marcello33:
Any news on this being merged?
It's so frustrating to only work with strings

bmoffatt:
This isn't a valid change, API Gateway will reject the response if the Body field isn't a string.

With this change, any function that tries to set Body to a non-string value, instead of getting their desired JSON, will instead see `{"message": "Internal Server Error"}`

With API Gateway access logs enabled, I see this error for the failed request: `The response from the Lambda function doesn't match the format that API Gateway expects. Lambda body contains the wrong type for field "body"` 

However, if you find marshaling to a json string to be cumbersome, you also have the option to return any object instead of the envelope struct `APIGatewayV2HTTPResponse`. 

from: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
> With the 2.0 format version, API Gateway can infer the response format for you. API Gateway makes the following assumptions if your Lambda function returns valid JSON and doesn't return a statusCode:
>   * isBase64Encoded is false
>   * statusCode is 200.
>   * content-type is application/json.
>   * body is the function's response.


here's the test resources I used:

----

SAM `template.yaml`
``` yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: >
  go-http-playground
  
  Sample SAM Template for go-http-playground

Globals:
  Function:
    MemorySize: 128
    Timeout: 5
    CodeUri: hello-world/
    Handler: hello-world
    Runtime: go1.x

Resources:
  Hello1:
    Type: AWS::Serverless::Function 
    Properties:
      Environment:
        Variables:
          HANDLER: "1"
      Events:
        Proxy:
          Type: HttpApi
          Properties:
            ApiId: !Ref HttpApi
            Path: /hello1
            Method: ANY

  Hello2:
    Type: AWS::Serverless::Function 
    Properties:
      Environment:
        Variables:
          HANDLER: "2"
      Events:
        Proxy:
          Type: HttpApi
          Properties:
            ApiId: !Ref HttpApi
            Path: /hello2
            Method: ANY

  Hello3:
    Type: AWS::Serverless::Function 
    Properties:
      Environment:
        Variables:
          HANDLER: "3"
      Events:
        Proxy:
          Type: HttpApi
          Properties:
            ApiId: !Ref HttpApi
            Path: /hello3
            Method: ANY

  HttpApi:
    Type: AWS::Serverless::HttpApi
    Properties:
      AccessLogSettings:
        DestinationArn: !GetAtt AccessLogs.Arn
        Format: $context.requestId $context.path $context.integration.integrationStatus $context.integrationErrorMessage

  AccessLogs:
    Type: AWS::Logs::LogGroup

Outputs:
  Hello1API:
    Value: !Sub "https://${HttpApi}.execute-api.${AWS::Region}.amazonaws.com/hello1"
  Hello2API:
    Value: !Sub "https://${HttpApi}.execute-api.${AWS::Region}.amazonaws.com/hello2"
  Hello3API:
    Value: !Sub "https://${HttpApi}.execute-api.${AWS::Region}.amazonaws.com/hello3"
```

`hello-world/handler.go`
``` go
package main

import (
	"encoding/json"
	"net/http"
	"os"

	"github.com/aws/aws-lambda-go/lambda"
)

const HandlerEnvVar = "HANDLER"

// CurrentAPIGatewayV2HTTPResponse is copy-paste of APIGatewayV2HTTPResponse from github.com/aws/aws-lambda-go/events
type CurrentAPIGatewayV2HTTPResponse struct {
	StatusCode        int                 `json:"statusCode"`
	Headers           map[string]string   `json:"headers"`
	MultiValueHeaders map[string][]string `json:"multiValueHeaders"`
	Body              interface{}         `json:"body"`
	IsBase64Encoded   bool                `json:"isBase64Encoded,omitempty"`
	Cookies           []string            `json:"cookies"`
}

// ProposedAPIGatewayV2HTTPResponse reflects the model changed proposed in PR #373
type ProposedAPIGatewayV2HTTPResponse struct {
	StatusCode        int                 `json:"statusCode"`
	Headers           map[string]string   `json:"headers"`
	MultiValueHeaders map[string][]string `json:"multiValueHeaders"`
	Body              interface{}         `json:"body"`
	IsBase64Encoded   bool                `json:"isBase64Encoded,omitempty"`
	Cookies           []string            `json:"cookies"`
}

// Hello returns our handler's response, the caller will handle proxying it to APIGateway
func Hello() interface{} {
	return struct {
		Handler  string
		Messages string
	}{
		os.Getenv(HandlerEnvVar),
		"Hello World!!!",
	}
}

// ProposedV2Handler demonstrates the proposed changes to the envelope. API Gateway rejects this response!!
func ProposedV2Handler() (ProposedAPIGatewayV2HTTPResponse, error) {
	return ProposedAPIGatewayV2HTTPResponse{
		StatusCode: http.StatusOK,
		Body:       Hello(),
	}, nil
}

// CurrentV2Handler demonstrates the construction of an API Gateway HttpApi response envelope containing stringifyed JSON
func CurrentV2Handler() (CurrentAPIGatewayV2HTTPResponse, error) {
	responseBytes, _ := json.Marshal(Hello())
	return CurrentAPIGatewayV2HTTPResponse{
		StatusCode: http.StatusOK,
		Body:       string(responseBytes),
	}, nil
}

// AlternativeV2Handler demonstrates that with APIGateway's HttpApi, Lambda functions can now return json without the proxy envelope
func AlternativeV2Handler() (interface{}, error) {
	return Hello(), nil
}

func main() {
	handlers := map[string]lambda.Handler{
		"1": lambda.NewHandler(CurrentV2Handler),
		"2": lambda.NewHandler(ProposedV2Handler),
		"3": lambda.NewHandler(AlternativeV2Handler),
	}
	lambda.StartHandler(handlers[os.Getenv(HandlerEnvVar)])
}
```

``` bash
$ for i in 1 2 3; do echo "handler=$i"; curl -s https://<redacted>.execute-api.us-west-2.amazonaws.com/hello$i | jq; done
handler=1
{
  "Handler": "1",
  "Messages": "Hello World!!!"
}
handler=2
{
  "message": "Internal Server Error"
}
handler=3
{
  "Handler": "3",
  "Messages": "Hello World!!!"
}
```

PR #378, opened 2021-05-22
Adding ECR Image Scan Event

alvinsiew:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/events/ecr_scan.go
+++ b/events/ecr_scan.go
@@ -1,26 +1,26 @@
 package events
 
-type EcrScanEvent struct {
-	Version    string     `json:"version"`
-	ID         string     `json:"id"`
-	DetailType string     `json:"detail-type"`
-	Source     string     `json:"source"`
-	Time       string     `json:"time"`
-	Region     string     `json:"region"`
-	Resources  []string   `json:"resources"`
-	Account    string     `json:"account"`
-	Detail     DetailType `json:"detail"`
+type ECRScanEvent struct {
+	Version    string                 `json:"version"`
+	ID         string                 `json:"id"`
+	DetailType string                 `json:"detail-type"`
+	Source     string                 `json:"source"`
+	Time       string                 `json:"time"`
+	Region     string                 `json:"region"`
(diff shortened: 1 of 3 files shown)


events/ecr_scan.go
```diff
@@ -0,0 +1,30 @@
+package events
+
+type EcrScanEvent struct {
```
bmoffatt, line 3: initialisms&acronyms should have consistent casing. eg: `ECRScanEvent`

events/ecr_scan.go
```diff
@@ -0,0 +1,30 @@
...
+	DetailType string     `json:"detail-type"`
+	Source     string     `json:"source"`
+	Time       string     `json:"time"`
+	Region     string     `json:"region"`
+	Resources  []string   `json:"resources"`
+	Account    string     `json:"account"`
+	Detail     DetailType `json:"detail"`
+}
+
+type DetailType struct {
```
bmoffatt, line 15: need to namespace sub-types, since the package events package is so broad. So this should be `ECRScanEventDetailType`

events/ecr_scan.go
```diff
@@ -0,0 +1,30 @@
...
+	ScanStatus            string                    `json:"scan-status"`
+	RepositoryName        string                    `json:"repository-name"`
+	FindingSeverityCounts FindingSeverityCountsType `json:"finding-severity-counts"`
+	ImageDigest           string                    `json:"image-digest"`
+	ImageTags             []string                  `json:"image-tags"`
+}
+
+type FindingSeverityCountsType struct {
```
bmoffatt, line 23: ditto, type should be named like `ECRScanEventFindingSeverityCounts`

bmoffatt (changes requested):
comments are on the type names only, structure and tests look correct when compared with the docs: https://docs.aws.amazon.com/AmazonECR/latest/userguide/ecr-eventbridge.html

bmoffatt (approved):
LGTM!

PR #383, opened 2021-07-20
bump codecov/codecov-action@v2

shogo82148:
*Issue #, if available:*
N/A

*Description of changes:*

`codecov/codecov-action@v2` is released and v2 is now deprecated.

> https://github.com/codecov/codecov-action#%EF%B8%8F--deprecration-of-v1
> Deprecration of v1
> On February 1, 2022, this version will be fully sunset and no longer function
> 
> Due to the deprecation of the underlying bash uploader, the Codecov GitHub Action has released v2 which will use the new uploader. You can learn more about our deprecation plan and the new uploader on our blog.
> 
> We will be restricting any updates to the v1 Action to security updates and hotfixes.

We should migrate from v1 to v2.


.github/workflows/tests.yml
```diff
@@ -29,7 +29,9 @@ jobs:
...
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v1
-        if: matrix.go == '1.16'
+        uses: codecov/codecov-action@v2
         with:
           file: ./coverage.txt
+          env_vars: GO
+        env:
+          GO: ${{ matrix.go }}
```
bmoffatt, line 37: Does this mean that each PR in future may be N codecov comments? I guess it'll be obvious once merged

bmoffatt (approved):
let's see how it goes!

PR #385, opened 2021-07-22
Add bootstarpServers and headers to Kafka event

ianpyw:
This change update the Kafka event to add `bootstrapbrokers` and record headers.


events/kafka_test.go
```diff
@@ -20,21 +20,25 @@ func TestKafkaEventMarshaling(t *testing.T) {
...
 	}
-
-	// 3. serialize to JSON
-	outputJson, err := json.Marshal(inputEvent)
-	if err != nil {
-		t.Errorf("could not marshal event. details: %v", err)
-	}
-
-	// 4. check result
-	assert.JSONEq(t, string(inputJson), string(outputJson))
```
ianpyw, line 37: The Kafka header is a list of map between string and []byte. 
According to [Marshal JSON](https://pkg.go.dev/encoding/json#Marshal)
```
Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value.
```
This assertion won't be true anymore.

zsombor-balogh (approved):
Lgtm

smirnoal (approved):
lgtm

Issue #372, opened 2021-04-16
Payload version 2 event AuthResponse isAuthorized


kaihendry:
**Describe the solution you'd like**
Struct for the Payload 2.0 type https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html

**Describe alternatives you've considered**
Hard coding the response

![image](https://user-images.githubusercontent.com/765871/114963941-60262280-9ea0-11eb-84f5-082a207322a5.png)

kaihendry:
Furthermore, I'm unable to find a struct that matches up to https://s.natalian.org/2021-04-16/payload.json

Issue #369, opened 2021-03-14
HTTP API (API Gateway v2) support for Payload Format v2.0


abelperezok:
**Is your feature request related to a problem? Please describe.**

One of the possible integrations with Lambda is HTTP APIs  aka API Gateway v2, the event object representation is missing (or at least I couldn't find it). It should describe the Payload Format v2.0.  The lack of it makes it difficult to integrate with such event.  More information in [AWS doics](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html#2.0)

**Describe the solution you'd like**
A new struct similar to APIGatewayProxyRequest which is tailored to the Payload Format v2.0. An existing implementation has been released for .NET runtime using the name [APIGatewayHttpApiV2ProxyRequest](https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayHttpApiV2ProxyRequest.cs) 

**Describe alternatives you've considered**
The alternative is to create my own struct or force to use Payload Format v1.0. It would be good to have the specific struct available as part of this package. 

**Additional context**
If this is something  that might be considered low priority or the team is too busy focusing on something more important, I'll be happy to prepare a pull request to fix this.

redaLaanait:
@abelperezok which version of the **aws-lambda-go** are you using? It seems that **API Gateway V2** is supported since v1.16.0:

[https://github.com/aws/aws-lambda-go/pull/274/files](https://github.com/aws/aws-lambda-go/pull/274/files)

abelperezok:
@redaLaanait My mistake, I was using a much older version, 1.13. Thank you very much for pointing it out. I suspect the template that comes with AWS SAM CLI when running `sam init` needs some updates.

PR #386, opened 2021-08-10
Add API Gateway V2 custom authorizer simple response type

dirknilius:
*Issue #369, #372:*

*Description of changes:*

This PR adds the missing type for API Gateway V2 custom authorizer simple response. See: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html#http-api-lambda-authorizer.payload-format-response

### DIFF
--- a/events/apigw.go
+++ b/events/apigw.go
@@ -288,7 +288,7 @@ type APIGatewayCustomAuthorizerResponse struct {
 // APIGatewayV2CustomAuthorizerSimpleResponse represents the simple format of an API Gateway V2 authorization response.
 type APIGatewayV2CustomAuthorizerSimpleResponse struct {
 	IsAuthorized bool                   `json:"isAuthorized"`
-	Context      map[string]interface{} `json:"context"`
+	Context      map[string]interface{} `json:"context,omitempty"`
 }
 
 // APIGatewayCustomAuthorizerPolicy represents an IAM policy
--- a/events/apigw.go
+++ b/events/apigw.go
@@ -285,6 +285,12 @@ type APIGatewayCustomAuthorizerResponse struct {
 	UsageIdentifierKey string                           `json:"usageIdentifierKey,omitempty"`
 }
 
+// APIGatewayV2CustomAuthorizerSimpleResponse represents the simple format of an API Gateway V2 authorization response.
+type APIGatewayV2CustomAuthorizerSimpleResponse struct {
+	IsAuthorized bool                   `json:"isAuthorized"`
+	Context      map[string]interface{} `json:"context"`
+}
+
 // APIGatewayCustomAuthorizerPolicy represents an IAM policy
 type APIGatewayCustomAuthorizerPolicy struct {
 	Version   string
--- a/events/apigw_test.go
+++ b/events/apigw_test.go
@@ -186,6 +186,7 @@ func TestAPIGatewayV2CustomAuthorizerSimpleResponseMarshalling(t *testing.T) {
 		{"defaults", APIGatewayV2CustomAuthorizerSimpleResponse{}, `{"isAuthorized":false}`},
 		{"without context", APIGatewayV2CustomAuthorizerSimpleResponse{IsAuthorized: true}, `{"isAuthorized":true}`},
 		{"context is nil", APIGatewayV2CustomAuthorizerSimpleResponse{Context: nil}, `{"isAuthorized":false}`},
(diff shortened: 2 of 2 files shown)


events/apigw.go
```diff
@@ -285,6 +285,12 @@ type APIGatewayCustomAuthorizerResponse struct {
 	UsageIdentifierKey string                           `json:"usageIdentifierKey,omitempty"`
 }
 
+// APIGatewayV2CustomAuthorizerSimpleResponse represents the simple format of an API Gateway V2 authorization response.
+type APIGatewayV2CustomAuthorizerSimpleResponse struct {
+	IsAuthorized bool                   `json:"isAuthorized"`
+	Context      map[string]interface{} `json:"context"`
```
harrisonhjones, line 291: Did you get a chance to verify this works as expected with a context set and without it set (using the default value). I just wonder what happens if `Context` is nil (`null` in json) and/or an empty map (`{}` in json)

dirknilius: Yes, already using it in production for a while. Works with or without context. So the API accepts the input. But I could check what’s exactly generated and if it makes sense to add `omitempty`.

harrisonhjones: It might make sense and save just a few bytes over the long run so if you could check that would be great

dirknilius: Hey @harrisonhjones, I just added the tests. Pls take a look if I covered everything.

dirknilius: BTW, I also added the `omitempty` flag, as the results are more reasonable.

events/apigw_test.go
```diff
@@ -177,6 +177,35 @@ func TestApiGatewayCustomAuthorizerResponseMarshaling(t *testing.T) {
...
+		{"without context", APIGatewayV2CustomAuthorizerSimpleResponse{IsAuthorized: true}, `{"isAuthorized":true}`},
+		{"context is nil", APIGatewayV2CustomAuthorizerSimpleResponse{Context: nil}, `{"isAuthorized":false}`},
+		{"context is empty", APIGatewayV2CustomAuthorizerSimpleResponse{Context: map[string]interface{}{}}, `{"isAuthorized":false}`},
+		{"context with basic types", APIGatewayV2CustomAuthorizerSimpleResponse{Context: map[string]interface{}{"string": "value", "bool": true, "number": 1234}}, `{"isAuthorized":false,"context":{"string":"value","bool":true,"number":1234}}`},
+		{"context with array", APIGatewayV2CustomAuthorizerSimpleResponse{Context: map[string]interface{}{"array": []string{"value1", "value2"}}}, `{"isAuthorized":false,"context":{"array":["value1","value2"]}}`},
+		{"context with map", APIGatewayV2CustomAuthorizerSimpleResponse{Context: map[string]interface{}{"map": map[string]string{"key": "value"}}}, `{"isAuthorized":false,"context":{"map":{"key":"value"}}}`},
+		{"context with nil value", APIGatewayV2CustomAuthorizerSimpleResponse{Context: map[string]interface{}{"nil": nil}}, `{"isAuthorized":false,"context":{"nil":null}}`},
```
harrisonhjones, line 193: nit: these might be a bit easier to read if the various properties were on separate lines but I wouldn't block on it.

events/apigw_test.go
```diff
@@ -177,6 +177,35 @@ func TestApiGatewayCustomAuthorizerResponseMarshaling(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
harrisonhjones, line 197: IIRC you only need to do this if you are running things in parallel (which you actually could do here).

harrisonhjones (approved):
LGTM.

events/apigw_test.go
dirknilius: Sure thing.  Just applied the suggested change.

events/apigw_test.go
dirknilius: In most cases I would agree. But without pinning, it can lead to subtle bugs in certain cases. Like described in the `loop pointer` (https://github.com/kyoh86/looppointer) linter used by `golangci-lint`. In this case you are absolutely right. I was just following the principal of pinning all the time like the linter suggests. I would leave it in if you don't mind.

harrisonhjones: Yea. That's fine and makes sense.

harrisonhjones (approved):

chrispruitt:
This will be great! Can this get merged soon?

harrisonhjones:
@bmoffatt or someone else on the Lambda team would need to merge it in.

bmoffatt (approved):
LGTM!

PR #406, opened 2021-10-31
events: Adds Custom Resource event

masoudkarimif:
Adds **CloudFormation Custom Resource** event type.

Tests based on examples sourced from AWS documentation. Examples can be found [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html) and [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-requesttypes-update.html).


leepa:
Does this not already exist in the cfn folder?

masoudkarimif:
> Does this not already exist in the cfn folder?

You're right. Sorry, I missed that. I'll close this one. Thanks!

PR #408, opened 2021-11-12
Add `omitempty` to ALBTargetGroupRequest.Body

WhyNotHugo:
If a lambda returns `Body: nil`, then the ALB returns a 502.


bmoffatt:
Verified that ALB does as you say, 502 returned with json `{"statusCode": 200, "body": null}` but expect 200 returned when json is `{"statusCode": 200}`. The go struct isn't actually able to produce this, since the type is `string` rather than `*string`.

So... I think it's OK to represent in this model that `null` should never be written for this integration. As far as I can tell in my testing it's also harmless to drop the empty string body, ALB behaves the same way.

bmoffatt (approved):

bmoffatt:
> Verified that ALB does as you say, 502 returned with json `{"statusCode": 200, "body": null}` but expect 200 returned when json is `{"statusCode": 200}`. The go struct isn't actually able to produce this, since the type is `string` rather than `*string`.
> 
> So... I think it's OK to represent in this model that `null` should never be written for this integration. As far as I can tell in my testing it's also harmless to drop the empty string body, ALB behaves the same way.

I was referencing this issue recently in a discussion, and realized that the change modified the request, not the response - so the issue described of a 5xx on the response path still exists.

WhyNotHugo:
:facepalm: You're right.

PR #410, opened 2021-11-16
Create BatchItemFailure structs for DynamoDB, Kinesis, SQS

lyoung-confluent:
This change adds the _response_ structs for "reporting batch item failures" documented in:
* [DynamoDB: Reporting batch item failures
](https://docs.aws.amazon.com/lambda/latest/dg/with-ddb.html#services-ddb-batchfailurereporting)
* [Kinesis: Reporting batch item failures
](https://docs.aws.amazon.com/lambda/latest/dg/with-kinesis.html#services-kinesis-batchfailurereporting)
* [SQS: Reporting batch item failures](docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-sqs-batchfailurereporting)

This is implemented in the Java events SDK as: [aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/StreamsEventResponse.java ](https://github.com/aws/aws-lambda-java-libs/blob/311f97cc47b414633c62ee96d9df3e8404e14d69/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/StreamsEventResponse.java)

I don't know that adding a test here has any value, but happy to add one if you'd like.


bmoffatt:
Thanks! This looks applicable for SQS Partial Batch Failures too!
https://aws.amazon.com/about-aws/whats-new/2021/11/aws-lambda-partial-batch-response-sqs-event-source/
https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-sqs-batchfailurereporting

Though it can be used across all 3 event sources, I'm thinking to proactively namespace the structs. eg: define `KinesisEventResponse`, `DynamoDBEventResponse`, `SQSEventResponse` 

@carlzogh WDYT?

carlzogh:
> Though it can be used across all 3 event sources, I'm thinking to proactively namespace the structs. eg: define `KinesisEventResponse`, `DynamoDBEventResponse`, `SQSEventResponse`

Absolutely, thanks @bmoffatt - we've found it tricky in the Java events library to rename/cater for different event source types that seem similar at first, so accounting for those explicitly from the get-go is a good idea!

shawnkoon:
Doing a pulse check, would love to see `SQSEventResponse` implemented.

> Though it can be used across all 3 event sources, I'm thinking to proactively namespace the structs. eg: define KinesisEventResponse, DynamoDBEventResponse, SQSEventResponse

Is there a strategy on `BatchItemFailure` ex) if it should be name spaced to `{SQS|Kinesis|DynamoDB|Stream}BatchItemFailure` or just keep one `BatchItemFailure` somewhere? Happy to contribute

lyoung-confluent:
I split up the structs (and child item failure structs) into different structs per each service (as well as adding SQS) and updated the description/title accordingly. I kept everything in `streams.go` instead of individual files though.

bmoffatt (approved):

bmoffatt:
LGTM, thanks!

PR #415, opened 2021-12-07
add source-version to additional information in codebuild details

chrispruitt:
This adds the `source-version` to the CodeBuildEventAdditionalInformation struct. The value is the commit id when the source is github or bitbucket.

### DIFF
--- a/events/codebuild.go
+++ b/events/codebuild.go
@@ -109,6 +109,8 @@ type CodeBuildEventAdditionalInformation struct {
 
 	Source CodeBuildSource `json:"source"`
 
+	SourceVersion string `json:"source-version"`
+
 	Logs CodeBuildLogs `json:"logs"`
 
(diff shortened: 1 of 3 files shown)


bmoffatt:
LGTM, do you have a reference to the corresponding AWS Docs I can look at to verify? If not that's fine, just need to make some time to trigger the event myself :)

chrispruitt:
I don't. The only docs I found on the codebuild notification event, were only example JSONs that didn't include this field. But I found that it exists when actually consuming codebuild notifications.

https://docs.aws.amazon.com/codebuild/latest/userguide/sample-build-notifications.html

bmoffatt:
I was able to verify this by starting a build using the "Start build with overrides" flow in the console.

bmoffatt (approved):

Issue #413, opened 2021-11-29
Missing build-number in CodeBuildEventAdditionalInformation


guiminyao:
**Is your feature request related to a problem? Please describe.**
The [CodeBuildEventAdditionalInformation](https://github.com/aws/aws-lambda-go/blob/c481c6113857aba7c3b04eab56af78dbe970cfb1/events/codebuild.go#L97) does not have the field for BuildNumber.
It seems not presented in IN_PROGRESS event. but SUCCEEDED, FAILED, STOPPED events do have the data.

**Describe the solution you'd like**
add the field added to the struct.

> BuildNumber int `json:"build-number,omitempty"`

Thanks.

guiminyao:
The fix PR #417 has been merged.

PR #417, opened 2021-12-16
Add missing BuildNumber to CodeBuildEventAdditionalInformation

guiminyao:
*Issue #, if available:*
https://github.com/aws/aws-lambda-go/issues/413

*Description of changes:*
Add missing BuildNumber to CodeBuildEventAdditionalInformation struct, and update tests.
SUCCEEDED, FAILED, STOPPED events do have the piece of info, not presented in IN_PROGRESS event.

### DIFF
--- a/events/codebuild.go
+++ b/events/codebuild.go
@@ -103,6 +103,8 @@ type CodeBuildEventAdditionalInformation struct {
 
 	BuildComplete bool `json:"build-complete"`
 
+	BuildNumber int `json:"build-number,omitempty"`
+
 	Initiator string `json:"initiator"`
 
 	BuildStartTime CodeBuildTime `json:"build-start-time"`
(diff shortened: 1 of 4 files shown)


events/codebuild.go
```diff
@@ -103,6 +103,8 @@ type CodeBuildEventAdditionalInformation struct {
 
 	BuildComplete bool `json:"build-complete"`
 
+	BuildNumber CodeBuildNumber `json:"build-number,omitempty"`
```
bmoffatt, line 106: why a new type and custom serializer? Does the builtin `int` not work for this?

```BuildNumber int `json:"build-number,omitempty"` ```

guiminyao: The original value of build-number in CodeBuild JSON event is not presented as integer but float (e.g. 55.0), have to use a new type to leverage custom marshalling to convert it to `int`. Cheers.

bmoffatt: gotcha, makes sense to me

bmoffatt (approved):

Issue #433, opened 2022-03-24
Allow Error return customization


driverpt:
**Is your feature request related to a problem? Please describe.**
We have the need to customize Error responses, e.g.: Add an Error Code/Status Code. And we feel that it's kind of limiting to only have `messages.InvokeResponse_Error{}` as Error Return.

**Describe the solution you'd like**
Is it possible to make this customizable via Custom JSON Serialization ?

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

bmoffatt:
integrations that call Lambdas have had years to ossify the assumption that an error json will look something like this:
```
{
...
  "errorMessage": ...,
...
}
```

so the approach for this probably shouldn't be as simple as delegating `json.Marshal`, that feels like a footgun for a bunch of users if `errorMessage` could accidentally go missing. Somehow the `messages.InvokeResponse_Error{}` and your custom serialization result would have to be merged. Or perhaps the custom serialization result nested inside like:

```
type InvokeResponse_Error struct {
	Message    string                             `json:"errorMessage"`
	Type       string                             `json:"errorType"`
	StackTrace []*InvokeResponse_Error_StackFrame `json:"stackTrace,omitempty"`
	ShouldExit bool                               `json:"-"`
	// new field, populated with the result of `json.Marshal` if the handler error has a `MarshalJSON` receiver defined
	Custom json.RawMessage                        `json:"custom,omitempty"`
}

```

/cc @carlzogh 

If we went this this approach, we'd also wanna check that it behaves with the `go1.x` RPC client.

----

@driverpt Are you running your functions with `go1.x`, `provided`, or `provided.al2`? If your custom fields appeared nested in the existing error json, would that still fit your use case?

driverpt:
Our functions are running on `provided.al2`.
 Ideally it would have a `json.Marshall(err)`.

Is it possible to have a Global Flag to tell the handler to use Custom errors?

PR #444, opened 2022-05-18
Add capability to define optional handler behavior

bmoffatt:
*Issue #, if available:*

https://github.com/aws/aws-lambda-go/issues/177
https://github.com/aws/aws-lambda-go/pull/178#discussion_r266617767

*Description of changes:*

* add Option WithSetEscapeHTML to set this same named option on the underlying json encoder
* refactor implementations of ***WithContext functions to use the Option WithContext
* mark a bunch of not useful related stuff as deprecated so that they are hidden in the godocs

This adds an options pattern for allowing a function author to pick from one or more customizations to the default runtime behavior. The pattern can later be followed to implement other requests for changes to the default runtime behavior. Example feature request:  https://github.com/aws/aws-lambda-go/issues/433

### DIFF
--- a/lambda/handler.go
+++ b/lambda/handler.go
@@ -28,6 +28,14 @@ type handlerOptions struct {
 type Option func(*handlerOptions)
 
 // WithContext is a HandlerOption that sets the base context for all invocations of the handler.
+//
+// Usage:
+//  lambda.StartWithOptions(
+//   	func (ctx context.Context) (string, error) {
+//   		return ctx.Value("foo"), nil
+//   	},
+//   	lambda.WithContext(context.WithValue(context.Background(), "foo", "bar"))
+//  )
 func WithContext(ctx context.Context) Option {
 	return Option(func(h *handlerOptions) {
 		h.baseContext = ctx
@@ -35,6 +43,14 @@ func WithContext(ctx context.Context) Option {
 }
 
 // WithSetEscapeHTML sets the SetEscapeHTML argument on the underlying json encoder
+//
+// Usage:
+//  lambda.StartWithOptions(
+//  	func () (string, error) {
+//  		return "<html><body>hello!></body></html>", nil
+//  	},
+//  	lambda.WithSetEscapeHTML(true),
+//  )
 func WithSetEscapeHTML(escapeHTML bool) Option {
 	return Option(func(h *handlerOptions) {
 		h.jsonResponseEscapeHTML = escapeHTML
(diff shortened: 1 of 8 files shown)


lambda/handler.go
```diff
@@ -15,29 +17,36 @@ type Handler interface {
 	Invoke(ctx context.Context, payload []byte) ([]byte, error)
 }
 
-// lambdaHandler is the generic function type
-type lambdaHandler func(context.Context, []byte) (interface{}, error)
+type handlerOptions struct {
+	Handler
+	baseContext context.Context
+	escapeHTML  bool
```
carlzogh, line 23: If this is only relevant to response serialization, maybe we can make it explicit in the property name (eg. `escapeResponseHTML` or something along those lines)

carlzogh:
Thanks for this Bryan! It may be worth adding a little Usage section to describe how a user of the library would set some of these options

carlzogh (approved):

lambda/handler.go
harrisonhjones: I agree. I would even go a bit further and prefix/suffix these with something like `ResponseJSON` so it would be something like:

```
responseJSONEscapeHTML
responseJSONPrefix
responseJSONIndex
```

Alternatively you could have customers pass in a `newJSONEncoder` function which they could configure however they like (or use a different JSON library). It would have the following signature:

```
newJSONEncoder func(out io.Writer) error
```

That might be the most flexible. Could even throw in a `newJSONDecoder` so users could have complete control over the JSON part of the process.

bmoffatt: I did play around with the custom encoder idea before pushing this PR! However didn't come up with something that still elegantly worked with the stdlib json NewEncoder. My failed ideas looked something like this: https://play.golang.com/p/To4q1sKBbjm 

Something like this would work:
```
func WithEncoder(encoderFactory func(io.Writer) func(any) error) Option {
  return Option(func(h *handlerOption) {
    h.encoderFactory = encoderFactory
  })
}
// use example with stdlib json
encoderOption := WithEncoder(func(w io.Writer) func(any) error {
  return json.NewEncoder(w).Encode
})
```
But i'd want to mull on it a little more, and look at other popular encoders, to see if there's not something more ergonomic that can be figured out

harrisonhjones: yea, I personally like the `WithEncoder` pattern over having three new json-specific variables but either way ultimately works.

PR #451, opened 2022-07-11
Add ExtendedRequestId to APIGatewayProxyRequestContext

DanielBauman88:
It exists on APIGatewayWebsocketProxyRequestContext but not ExtendedRequestId.
It is however present on non-websocket requests and should be accessible by consumers of this package.

* REST https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html
* HTTP https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-logging-variables.html

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/events/apigw.go
+++ b/events/apigw.go
@@ -30,22 +30,23 @@ type APIGatewayProxyResponse struct {
 // APIGatewayProxyRequestContext contains the information to identify the AWS account and resources invoking the
 // Lambda function. It also includes Cognito identity information for the caller.
 type APIGatewayProxyRequestContext struct {
-	AccountID        string                    `json:"accountId"`
-	ResourceID       string                    `json:"resourceId"`
-	OperationName    string                    `json:"operationName,omitempty"`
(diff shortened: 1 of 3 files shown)


DanielBauman88:
I validated this with my rest api by deploying my go code and replacing this module with my changes. The extended request id is accessible as expected.

This was supported for websocket requests when this was (added)[https://github.com/aws/aws-lambda-go/pull/159] but this was never supported for non-websocket requests.

This functionality is important so that customers can long the extendedRequestId in lambda code - this will match what's logged in the APIGW execution logs.

bmoffatt (approved):
https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format validated against docs sample

bmoffatt:
Thanks!

DanielBauman88:
Thanks!

Issue #318, opened 2020-08-24
Way to execute code on shutdown?


mikeparker:
I want to maintain an event metrics client across multiple lambda invocations so the server does not get spammed. However, I need to be able to flush the queue on shutdown.

Does the AWS go lib have any way to execute code on shutdown? If I simply add code to the end of `main` (after the `lambda.Start`), or using `defer`, will it get executed?

From viewing the code it looks like the main method will finish if there is a runtime error, but I'm unsure what happens if the lambda gets shutdown - I could not see code for dealing with a shutdown message (I dont think it gets a `SIGTERM`?)

jlordiales:
Same question, very similar use case. I initially just put the flush call on a `defer` from the handler function but that's obviously not ideal.
cc @bmoffatt

jlordiales:
I tried something like this today to see if the signal was being sent to the container but it didn't work either:
```go
done := make(chan struct{})
go func() {
  log.Println("Listening for signals...")
  c := make(chan os.Signal, 1)
  signal.Notify(c, os.Interrupt, syscall.SIGTERM)

  <-c
  close(done)
}()

go lambda.Start(HandleRequest)

<-done
log.Println("shutting down")
```

The lambda still works fine and I can see the "Listening for signals" log entry in CloudWatch but never the "shutting down" one

bmoffatt:
Today, teardown of functions happens with SIGKILL. We're aware of other asks for a shutdown signal (like SIGTERM), and I've passed this issue along to the product team for reference.

skiarn:
Hello, what is the recommended way to achieve same result? Since os/signal cannot be used to catch SIGKILL`The signals SIGKILL and SIGSTOP may not be caught by a program, and therefore cannot be affected by this package.` read more here: https://golang.org/pkg/os/signal

ianic:
I made few experiments and observe two different behaviors. 
First is explained in this thread; there is no way to get SIGTERM and that is true for Lambda functions without extensions. But if you register an external extensions for a Lambda function than functions gets SIGTERM and has 500 ms to do useful work. It is killed after that timeout. 

I first observe it when playing with [extensions](https://github.com/aws-samples/aws-lambda-extensions/issues/63). 
Then I repeat the same with the Go code. Make s dummy Go Lambda function: 
``` Go
type MyEvent struct {
	Name string `json:"name"`
}

func HandleRequest(ctx context.Context, name MyEvent) (string, error) {
[...]

PR #457, opened 2022-07-25
Add WithEnableSIGTERM option 

bmoffatt:
*Issue #, if available:*

https://github.com/aws/aws-lambda-go/issues/318

*Description of changes:*

Adds an incomplete internal implementation of the extensions API, and uses it to add an option to opt-in to SIGTERM. This allows function authors to run code on shutdown without *also* having to attach an extension layer.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,7 @@ jobs:
       - run: go version
 
       - name: install lambda runtime interface emulator
-        run: curl -o /usr/local/bin/aws-lambda-rie https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/download/v1.6/aws-lambda-rie-x86
+        run: curl -o /usr/local/bin/aws-lambda-rie https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/download/v1.6/aws-lambda-rie-x86_64
       - run: chmod +x /usr/local/bin/aws-lambda-rie 
 
       - name: Check out code into the Go module directory
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,7 @@ jobs:
       - run: go version
 
       - name: install lambda runtime interface emulator
-        run: curl -L -o /usr/local/bin/aws-lambda-rie https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/download/v1.6/aws-lambda-rie-x86_64
+        run: curl -L -o /usr/local/bin/aws-lambda-rie https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/latest/download/aws-lambda-rie-x86_64
       - run: chmod +x /usr/local/bin/aws-lambda-rie 
 
       - name: Check out code into the Go module directory
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,6 +25,10 @@ jobs:
       
       - run: go version
 
+      - name: install lambda runtime interface emulator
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -25,11 +25,15 @@ jobs:
       
       - run: go version
 
+      - name: install lambda runtime interface emulator
+        run: curl -L -o /usr/local/bin/aws-lambda-rie https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/download/v1.6/aws-lambda-rie-x86_64
```
carlzogh, line 29: change `v1.6` to `latest` in order to always test against the most recent release - eg. URL from RIE readme:
```
https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/latest/download/aws-lambda-rie-x86_64
```

lambda/handler.go
```diff
@@ -23,6 +23,8 @@ type handlerOptions struct {
 	jsonResponseEscapeHTML   bool
 	jsonResponseIndentPrefix string
 	jsonResponseIndentValue  string
+	enableSIGTERM            bool
+	sigtermCallbacks         []func()
```
carlzogh, line 27: Do we need an excplicit `enableSIGTERM` flag + another for registering callbacks? Would users ever enable SIGTERM without configuring handlers, or is `enableSIGTERM` redundant?

If not, I suggest combining into a single configuration option eg. `registerCallbacksSIGTERM`

lambda/sigterm.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
carlzogh, line 1: nit: would it make sense to move to a sub-package `signals`?

.github/workflows/tests.yml
bmoffatt: will do 👍

lambda/handler.go
bmoffatt: Yeah the signal handlers can be setup anywhere within the program, including in libraries outside of the function author's control. I include an example of this in the test cases. For these passing additional callbacks into `WithEnableSIGTERM` is optional.

```
func init() {
 	// conventional SIGTERM callback
 	signaled := make(chan os.Signal, 1)
 	signal.Notify(signaled, syscall.SIGTERM)
 	go func() {
 		<-signaled
 		fmt.Println("I've been TERMINATED!")
 	}()

 }
```

lambda/sigterm.go
bmoffatt: This would require making the enableSIGTERM helper public, which would make it easy to mistakenly use the extension registration APIs before the first invoke, which I believe results in API errors.

.github/workflows/tests.yml
bmoffatt:
```
$ curl -L https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/download/latest/aws-lambda-rie-x86_64
Not Found
```

carlzogh: Oh oops no arch suffix for x86_64 apparently

https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/latest/download/aws-lambda-rie

lambda/handler.go
carlzogh: Understood, makes sense - thanks!

carlzogh (dismissed):

.github/workflows/tests.yml
bmoffatt: Actually it seems to be a different url ordering for latest

blahblah/latest/download/

Not 

blahblah/download/v1.6/

🤷‍♂️

carlzogh (approved):

Issue #463, opened 2022-08-19
Users unable to get get build-lambda-zip (Failed on MacOS, Windows 10 and 11)


duaneking:
Windows users can not get build-lambda-zip

Repro Steps
On a clean win10 or win11 machine (this repos on both of mine) open an admin console:

go get -v -u github.com/aws/aws-lambda-go/cmd/build-lambda-zip

Actual Result: <no output from command, no binaries created in any directory>

No binaries get downloaded or build, nothing ends up in any of the bin directories.

duaneking:
I have reproduced this on a brand new clean install of win11 as well as an older install of win10.

All attempts to get the tools needed to deploy simply fail with no error message or output of any kind.

No tools are added to any ./bin/ folder.

duaneking:
>go.exe get -v -x -u github.com/aws/aws-lambda-go/cmd/build-lambda-zip
# get https://proxy.golang.org/github.com/@v/list
# get https://proxy.golang.org/github.com/aws/aws-lambda-go/cmd/@v/list
# get https://proxy.golang.org/github.com/aws/@v/list
# get https://proxy.golang.org/github.com/aws/aws-lambda-go/cmd/build-lambda-zip/@v/list
# get https://proxy.golang.org/github.com/aws/aws-lambda-go/@v/list
# get https://proxy.golang.org/github.com/@v/list: 404 Not Found (0.202s)
# get https://proxy.golang.org/github.com/aws/@v/list: 404 Not Found (0.202s)
# get https://proxy.golang.org/github.com/aws/aws-lambda-go/@v/list: 200 OK (0.208s)
# get https://proxy.golang.org/github.com/aws/aws-lambda-go/cmd/@v/list: 404 Not Found (1.752s)
# get https://proxy.golang.org/github.com/aws/aws-lambda-go/cmd/build-lambda-zip/@v/list: 404 Not Found (1.879s)

duaneking:
Looks like dependencies are failing

duaneking:
Its important to note that this is on a brand new install.

No prior versions of anything go related are on the system because its a new box.

duaneking:
```

go.exe get -v -x -u github.com/aws/aws-lambda-go/cmd/build-lambda-zip
# get https://proxy.golang.org/github.com/jmespath/go-jmespath/@v/v0.4.0.mod
# get https://proxy.golang.org/github.com/aws/aws-sdk-go/@v/v1.44.80.mod
[...]

PR #464, opened 2022-08-20
use `go install` to install `build-lambda-zip`.

shogo82148:
*Issue #, if available:*

fixes https://github.com/aws/aws-lambda-go/issues/463

*Description of changes:*

go 1.18 removed the ability to use `go get` to install binaries.
we should use `go install` instead.


bmoffatt (approved):

bmoffatt:
Thanks for the update!

PR #470, opened 2022-11-08
Adding properties field to activemq.go

lohithpc1:
*Issue #, if available:*
Custom Properties Set by the customer are not being propagated to target function, as we don't have a properties field in here.
*Description of changes:*
Added properties field and also a unit test to see if we are getting right fields.


SukanyaHanumanthu:
LGTM

SukanyaHanumanthu (approved):

PR #475, opened 2022-12-21
Fix issue 377

### LINKED ISSUE #377: Custom context for function handlers result in panic

In the AWS documentation, the handler for the function is shown to have the following signature

* func ()
* func () error
* func (TIn) error
* func () (TOut, error)
* func (TIn) (TOut, error)
* func (context.Context) error
* func (context.Context, TIn) error
* func (context.Context) (TOut, error)
* func (context.Context, TIn) (TOut, error)

https://docs.aws.amazon.com/lambda/latest/dg/golang-handler.html

But it also says the following.

> The handler may take between 0 and 2 arguments. If there are two arguments, the first argument must implement context.Context.

----

The first argument in the signature of the function that takes two arguments is clearly `context.Context`, but the `panic` occurs when `customContext` is taken as an argument. 

For example, deploy and execute the following code to AWS Lambda.

```go
package main

import (
	"context"
	"log"

	"github.com/aws/aws-lambda-go/lambda"
)

func main() {
	lambda.Start(Handle)
}

func Handle(ctx customContext) {
	log.Println("hello")
	return
}

// customContext satisfies "context.Context"
type customContext struct {
	context.Context
}
```

When this Lambda is executed, a panic occurs as shown below.

```
reflect: Call using *context.valueCtx as type main.customContext: string
[
	{
		"path": "github.com/aws/aws-lambda-go@v1.24.0/lambda/errors.go",
		"line": 39,
		"label": "lambdaPanicResponse"
	},
	{
		"path": "github.com/aws/aws-lambda-go@v1.24.0/lambda/function.go",
		"line": 36,
		"label": "(*Function).Invoke.func1"
	},
	{
		"path": "runtime/panic.go",
		"line": 965,
		"label": "gopanic"
	},
	{
		"path": "reflect/value.go",
		"line": 406,
		"label": "Value.call"
	},
	{
		"path": "reflect/value.go",
		"line": 337,
		"label": "Value.Call"
	},
	{
		"path": "github.com/aws/aws-lambda-go@v1.24.0/lambda/handler.go",
		"line": 124,
		"label": "NewHandler.func1"
	},
	{
		"path": "github.com/aws/aws-lambda-go@v1.24.0/lambda/handler.go",
		"line": 24,
		"label": "lambdaHandler.Invoke"
	},
	{
		"path": "github.com/aws/aws-lambda-go@v1.24.0/lambda/function.go",
		"line": 64,
		"label": "(*Function).Invoke"
	},
	{
		"path": "reflect/value.go",
		"line": 476,
		"label": "Value.call"
	},
	{
		"path": "reflect/value.go",
		"line": 337,
		"label": "Value.Call"
	},
	{
		"path": "net/rpc/server.go",
		"line": 377,
		"label": "(*service).call"
	},
	{
		"path": "runtime/asm_amd64.s",
		"line": 1371,
		"label": "goexit"
	}
]
```

If we take `context.Context` as an argument, we need to make sure that the first argument of the function signature is `context.Context` explicitly.

https://github.com/aws/aws-lambda-go/blob/5d6413264f82afaa7587b56e3f09f1422c898c68/lambda/handler.go#L48-L50

The above implementation only checks if the argument satisfies `context.Context`, so if a custom context that satisfies `context.Context` is the first argument, it cannot be checked by the above validation.

I wasn't sure if it was correct to explicitly accept only `context.Context` as a signature, or if we will accept a custom context satisfying `context.Context` as a function signature.

**Additional context**

Nothing.

shogo82148:
*Issue #, if available:*

fixes #377

*Description of changes:*

rework of https://github.com/aws/aws-lambda-go/pull/365, but it just accepts `context.Context` as a first argument.

see https://github.com/aws/aws-lambda-go/pull/365#pullrequestreview-1225172657

### DIFF
--- a/lambda/handler.go
+++ b/lambda/handler.go
@@ -99,20 +99,33 @@ func WithEnableSIGTERM(callbacks ...func()) Option {
 	})
 }
 
-func validateArguments(handler reflect.Type) (bool, error) {
-	handlerTakesContext := false
-	if handler.NumIn() > 2 {
-		return false, fmt.Errorf("handlers may not take more than two arguments, but handler takes %d", handler.NumIn())
-	} else if handler.NumIn() > 0 {
+// handlerTakesContext returns whether the handler takes a context.Context as its first argument.
+func handlerTakesContext(handler reflect.Type) (bool, error) {
+	switch handler.NumIn() {
+	case 0:
+		return false, nil
+	case 1:
 		contextType := reflect.TypeOf((*context.Context)(nil)).Elem()
 		argumentType := handler.In(0)
-		handlerTakesContext = argumentType.Implements(contextType)
-		if handler.NumIn() > 1 && !handlerTakesContext {
+		if argumentType.Kind() != reflect.Interface {
+			return false, nil
+		}
+		if argumentType.NumMethod() == 0 {
+			return false, nil
+		}
+		if !contextType.Implements(argumentType) || !argumentType.Implements(contextType) {
+			return false, fmt.Errorf("handler takes an interface, but it is not context.Context: %q", argumentType.Name())
+		}
+		return true, nil
+	case 2:
+		contextType := reflect.TypeOf((*context.Context)(nil)).Elem()
+		argumentType := handler.In(0)
+		if argumentType.Kind() != reflect.Interface || !contextType.Implements(argumentType) || !argumentType.Implements(contextType) {
 			return false, fmt.Errorf("handler takes two arguments, but the first is not Context. got %s", argumentType.Kind())
 		}
+		return true, nil
 	}
-
-	return handlerTakesContext, nil
+	return false, fmt.Errorf("handlers may not take more than two arguments, but handler takes %d", handler.NumIn())
 }
 
 func validateReturns(handler reflect.Type) error {
@@ -198,7 +211,7 @@ func reflectHandler(handlerFunc interface{}, h *handlerOptions) Handler {
 		return errorHandler(fmt.Errorf("handler kind %s is not %s", handlerType.Kind(), reflect.Func))
 	}
 
-	takesContext, err := validateArguments(handlerType)
+	takesContext, err := handlerTakesContext(handlerType)
 	if err != nil {
 		return errorHandler(err)
 	}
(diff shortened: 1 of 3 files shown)


lambda/handler.go
```diff
@@ -99,20 +99,33 @@ func WithEnableSIGTERM(callbacks ...func()) Option {
...
+		return false, nil
+	case 1:
 		contextType := reflect.TypeOf((*context.Context)(nil)).Elem()
 		argumentType := handler.In(0)
-		handlerTakesContext = argumentType.Implements(contextType)
-		if handler.NumIn() > 1 && !handlerTakesContext {
+		if argumentType.Kind() != reflect.Interface {
+			return false, nil
+		}
+		if argumentType.NumMethod() == 0 {
```
bmoffatt, line 113: This check doesn't appear in the `case 2:` branch, can it be removed?

lambda/handler.go
```diff
@@ -99,20 +99,33 @@ func WithEnableSIGTERM(callbacks ...func()) Option {
...
+			return false, nil
+		}
+		if !contextType.Implements(argumentType) || !argumentType.Implements(contextType) {
+			return false, fmt.Errorf("handler takes an interface, but it is not context.Context: %q", argumentType.Name())
```
bmoffatt, line 117: Wondering if there's a case where this isn't an error. The trival case is `any/interface{}`, where the `.NumMethod() == 0` check above makes sense. Wondering if the current version of the code panics if the first argument is `error`, or some other interface with one or more methods

lambda/handler.go
bmoffatt: I'm guessing this check is to keep handlers like `func(event any) {}` valid. If that's the case, it's worth adding a comment!

bmoffatt:
LGTM! Let me know if I understood the `.NumMethod() == 0` check correctly and than I can merge this in.

lambda/handler.go
shogo82148: yes.
If the first argument is an interface, the handler may be `func(event any)` or `func(ctx context.Context)`.
We should accept both.

bmoffatt (approved):

PR #477, opened 2022-12-26
Add HandlerFunc generics support

FlowingSPDG:
*Description of changes:*
Add generics for `HandlerFunc` interface for type-safe `StartHandlerFunc`

### DIFF
--- a/lambda/entry_generic_test.go
+++ b/lambda/entry_generic_test.go
@@ -21,7 +21,7 @@ func TestStartHandlerFunc(t *testing.T) {
 	}
 
 	f := func(context.Context, any) (any, error) { return 1, nil }
-	StartHandlerFunc(f)
+	StartHandlerFunc[any, any](f)
 
 	assert.Equal(t, "expected AWS Lambda environment variables [_LAMBDA_SERVER_PORT AWS_LAMBDA_RUNTIME_API] are not defined", actual)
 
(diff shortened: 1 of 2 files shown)


lambda/entry_generic_test.go
```diff
@@ -21,7 +21,7 @@ func TestStartHandlerFunc(t *testing.T) {
 	}
 
 	f := func(context.Context, any) (any, error) { return 1, nil }
-	StartHandlerFunc(f)
+	StartHandlerFunc[any, any](f)
```
bmoffatt, line 24: this is a breaking change, as the go compiler isn't smart enough to infer the types when more than one variant is added to the interface. See prior discussion: https://github.com/aws/aws-lambda-go/pull/468#discussion_r1034015336

bmoffatt:
Not merging this one, the requirement to specify the in/out types breaks existing users.

lambda/entry_generic_test.go
FlowingSPDG: I see, thank you for considering this @bmoffatt  !
I am still wondering if we can add this changes by adding another function (let's say `StartTypedHandlerFunc` for now)that accepts "any" type and it calls `StartHandlerFunc[any, any](f)`.
There will be no type-safe though, but user does not need to specify types for `StartHandlerFunc` . instead, they may want to use new `StartHandlerTypedFunc` that supports generics with type parameter.

Sorry my English is not good enough to explain this, hope it works!

PR #481, opened 2023-02-13
Added ECRImageActionEvent and test

nickryand:
*Issue #, if available:*

*Description of changes:*
Added ECR Image Action event support: https://docs.aws.amazon.com/AmazonECR/latest/userguide/ecr-eventbridge.html

### DIFF
--- a/events/ecr_image_action.go
+++ b/events/ecr_image_action.go
@@ -6,7 +6,7 @@ type ECRImageActionEvent struct {
 	DetailType string                        `json:"detail-type"`
 	Source     string                        `json:"source"`
 	Account    string                        `json:"account"`
-	Time       string                        `json:"time"`
+	Time       time.Time                     `json:"time"`
 	Region     string                        `json:"region"`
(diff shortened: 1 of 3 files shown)


events/ecr_image_action.go
```diff
@@ -0,0 +1,21 @@
+package events
+
+type ECRImageActionEvent struct {
+	Version    string                        `json:"version"`
+	ID         string                        `json:"id"`
+	DetailType string                        `json:"detail-type"`
+	Source     string                        `json:"source"`
+	Account    string                        `json:"account"`
+	Time       string                        `json:"time"`
```
bmoffatt, line 9: I think this can be `time.Time`. There's precedent in other eventbridge modeled events we have in this project. ex: https://github.com/aws/aws-lambda-go/blob/main/events/codedeploy.go#L23-L80

bmoffatt (approved):
LGTM! Thanks for the updates!

Issue #500, opened 2023-04-09
Handler design for streaming responses


aidansteele:
AWS recently released support for [streaming responses](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/) from Lambda functions. I can see that work is currently underway to add this to the Go SDK in #494, #495. It might be too late, but I'd like to raise an alternative API design for your consideration.

**Is your feature request related to a problem? Please describe.**
Right now it looks like streaming responses will be supported by having a handler function return an `io.Reader`. I think that this could potentially make the implementation of streaming-response Lambda functions more error-prone and provide less control over the streamed output than using an `io.Writer`. 

My understanding is that a streaming response handler ~today should be implemented like this:

```go
func handleReturningReader(ctx context.Context, input json.RawMessage) (io.Reader, error) {
	pr, pw := io.Pipe()

	go func() {
		// if this function panics, the entire process crashes. it can't be caught by
		// callBytesHandlerFunc like a panic in handleReturningReader would be. i suspect
		// a lot of users will be affected by this.

		for idx := 0; idx < 10; idx++ {
			fmt.Fprintln(pw, "hello world!")
			time.Sleep(time.Second)
		}
		
		if err := doSomething(); err != nil {
			// all mid-stream errors need to be funneled through the pipe. that's not
			// the end of the world, but how many users will know how to do this? it's
			// not as convenient as returning an error
			pw.CloseWithError(err)
		}

		// finally, if the user forgets to close the pipe (e.g. maybe they returned early)
		// then the invocation will just hang until it times out. the docs can advise
		// that the pipe must always be closed, but it's another opportunity for error
		pw.Close()
	}()

	return pr, nil
}
```

**Describe the solution you'd like**
Consider a handler signature like this:

```go
func handleReceivingWriter(ctx context.Context, output io.Writer, input json.RawMessage) error {
	// any panics in this function will be caught by the invoke loop and turned into an error
	// response like usual

	for idx := 0; idx < 10; idx++ {
		// the user has control over the size of each chunk in the chunked transfer-encoding.
		// this can simplify the usage of InvokeWithResponseStream on the client side
		fmt.Fprintln(output, "hello world!")
		time.Sleep(time.Second)
	}

	if err := doSomething(); err != nil {
		// error handling is the normal if err != nil { return nil } convention
		return err
	}

	return nil
}
```

[...]

PR #503, opened 2023-04-18
Support http.Handler for RESPONSE_STREAM Lambda Function URLs

bmoffatt:
*Issue #, if available:*

#500 

*Description of changes:*

Introduces a new package `github.com/aws/aws-lambda-go/lambdaurl` implementing an `http.Handler` based programming model for Lambda Function URLs configuered with `InvokeMode: RESPONSE_STREAM`

Decided follow the approach we chose for the `github.com/aws/aws-lambda-go/cfn` package. That package exposes a `Wrap` function that handles the translation of CloudFormation's unique programming model into the request/response+error model supported by `github.com/aws/aws-lambda-go/lambda.Start`. 

API Gateway (V1/V2), ALB Targets, and `InvokeMode: BUFFERED`, are explicitly not supported.
1. Today, it's not possible to tell within the function which `InvokeMode` the request came in with. 
2. Supporting `InvokeMode: BUFFERED` has a diminishing benefit - there's no reason I know of to choose `BUFFERED` for new applications
3. Supporting API Gateway (V1/V2), ALB Targets, is more scope than I can chew off in this PR, and has a diminishing benefit given that both can be configured to proxy to a Lambda Function URL

### DIFF
--- a/lambdaurl/http_handler.go
+++ b/lambdaurl/http_handler.go
@@ -93,7 +93,7 @@ func Wrap(handler http.Handler) func(context.Context, *events.LambdaFunctionURLR
 	}
 }
 
-// Start converts wraps a http.Handler and calls lambda.StartHandlerFunc
+// Start wraps a http.Handler and calls lambda.StartHandlerFunc
 // Only supports:
 //   - Lambda Function URLs configured with `InvokeMode: RESPONSE_STREAM`
 //   - Lambda Functions using the `provided` or `provided.al2` runtimes.
--- a/lambdaurl/http_handler.go
+++ b/lambdaurl/http_handler.go
@@ -48,7 +48,7 @@ func RequestFromContext(ctx context.Context) (*events.LambdaFunctionURLRequest, 
 
 // Wrap converts an http.Handler into a lambda request handler.
 // Only Lambda Function URLs configured with `InvokeMode: RESPONSE_STREAM` are supported with the returned handler.
-// The response body of the handler will conform the the content-type `application/vnd.awslambda.http-integration-response`
+// The response body of the handler will conform to the content-type `application/vnd.awslambda.http-integration-response`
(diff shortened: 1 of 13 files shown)


lambdaurl/http_handler.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		for k, v := range request.Headers {
+			httpRequest.Header.Add(k, v)
+		}
+		status := make(chan int) // Signals when it's OK to start returning the response body to Lambda
+		header := http.Header{}
+		r, w := io.Pipe()
+		go func() {
+			defer close(status)
+			defer w.Close() // TODO: recover and CloseWithError the any panic value once the runtime API client supports plumbing fatal errors through the reader
```
bmoffatt, line 75: https://github.com/aws/aws-lambda-go/issues/504

lambdaurl/http_handler.go
```diff
@@ -0,0 +1,103 @@
...
+
+// RequestFromContext returns the *events.LambdaFunctionURLRequest from a context.
+func RequestFromContext(ctx context.Context) (*events.LambdaFunctionURLRequest, bool) {
+	req, ok := ctx.Value(requestContextKey{}).(*events.LambdaFunctionURLRequest)
+	return req, ok
+}
+
+// Wrap converts an http.Handler into a lambda request handler.
+// Only Lambda Function URLs configured with `InvokeMode: RESPONSE_STREAM` are supported with the returned handler.
+// The response body of the handler will conform the the content-type `application/vnd.awslambda.http-integration-response`
```
aidansteele, line 51:
```suggestion
// The response body of the handler will conform to the content-type `application/vnd.awslambda.http-integration-response`
```

lambdaurl/http_handler.go
bmoffatt: Will get back to this in the follow-up for all returned writers. As for now allowing the panic to crash the process and log is the "most correct" thing do do, at least compared to swallowing it without any logging / reporting!

Issue #422, opened 2022-01-11
[BUG] Incorrect unmarshalling self-hosted Kafka event


LKay:
We have a lambda function which is hooked up and listens to both MSK and self-hosted Kafka instances to process incoming messages. While in case is MSK the event source mapping works fine, for self-hosted event source mapping SDK panics with error as follow:
```json
{
    "errorMessage": "json: cannot unmarshal number -23 into Go struct field KafkaRecord.records.headers of type uint8",
    "errorType": "UnmarshalTypeError"
}
```

This seems like some type mismatch (`uint8`) so it can't properly unmarshal negative numbers coming in the lambda event payload.

We had opened AWS support case and worked with them to some degree but unfortunately they couldn't provide us much context so we did debugging ourselves. 

The issue seems to be present in the versions `1.26+` of the Lambda SDK, and downgrading to 1.25 or below doesn't cause this unmarshal error to appear. The main different is the presence of the `headers` field in the struct introduced through https://github.com/aws/aws-lambda-go/pull/385/files#diff-cf0e5ac9714618f1821f8694da87d7af4579e0eafed210e8dbdcc5977bf5a69eR20. This might also be related to some previous issue https://github.com/aws/aws-lambda-go/issues/390.

As mentioned above this only applies to self-hosted event mapping, not MSK events. 

Environment:

- Go: `1.17`
- AWS LAmbda SDK Go: `1.26`, `1.27`, `1.28`
- Kafka version (self-hosted): `2.7.0`

bmoffatt:
Ah so the test data was correct, but the coverage that would've warned of the type mismatch was removed https://github.com/aws/aws-lambda-go/pull/385/files#r675441134

PR #502, opened 2023-04-16
feat: add ECSContainerInstanceEvent

wafuwafu13:
*Issue #, if available:*

*Description of changes:*
ref: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs_cwe_events.html#ecs_container_instance_events

I verified in my environment.

<details>
<summary>Lambda code</summary>

```go
package main

import (
	"context"
	"encoding/json"
	"fmt"
	"time"

	"github.com/aws/aws-lambda-go/lambda"
)

type ECSContainerInstanceEvent struct {
	Version    string                              `json:"version"`
	ID         string                              `json:"id"`
	DetailType string                              `json:"detail-type"`
	Source     string                              `json:"source"`
	Account    string                              `json:"account"`
	Time       time.Time                           `json:"time"`
	Region     string                              `json:"region"`
	Resources  []string                            `json:"resources"`
	Detail     ECSContainerInstanceEventDetailType `json:"detail"`
}

type ECSContainerInstanceEventDetailType struct {
	AgentConnected       bool        `json:"agentConnected"`
	Attributes           []Attribute `json:"attributes"`
	ClusterArn           string      `json:"clusterArn"`
	ContainerInstanceArn string      `json:"containerInstanceArn"`
	EC2InstanceID        string      `json:"ec2InstanceId"`
	RegisteredResources  []Resource  `json:"registeredResources"`
	RemainingResources   []Resource  `json:"remainingResources"`
	Status               string      `json:"status"`
	Version              int         `json:"version"`
	VersionInfo          VersionInfo `json:"versionInfo"`
	UpdatedAt            time.Time   `json:"updatedAt"`
}

type Attribute struct {
	Name string `json:"name"`
}

type Resource struct {
	Name           string    `json:"name"`
	Type           string    `json:"type"`
	IntegerValue   int       `json:"integerValue,omitempty"`
	StringSetValue []*string `json:"stringSetValue,omitempty"`
}

type VersionInfo struct {
	AgentHash     string `json:"agentHash"`
	AgentVersion  string `json:"agentVersion"`
	DockerVersion string `json:"dockerVersion"`
}

func (r Resource) MarshalJSON() ([]byte, error) {
	type Alias Resource
	aux := struct {
		StringSetValue json.RawMessage `json:"stringSetValue,omitempty"`
		Alias
	}{
		Alias: (Alias)(r),
	}

	if r.StringSetValue != nil {
		b, err := json.Marshal(r.StringSetValue)
		if err != nil {
[...]

PR #506, opened 2023-04-24
Changing Headers definition from byte to int8 for KafkaRecord

milovacb:
*Description of changes:*

The `Headers` field needs to reflect value range of `byte` data type in [the Java library](https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/KafkaEvent.java#L45).

The unit test example is consistent with public docs here: https://docs.aws.amazon.com/lambda/latest/dg/with-kafka.html#smaa-sample-event

### DIFF
--- a/events/kafka.go
+++ b/events/kafka.go
@@ -2,6 +2,10 @@
 
 package events
 
+import (
+	"encoding/json"
+)
+
 type KafkaEvent struct {
 	EventSource      string                   `json:"eventSource"`
 	EventSourceARN   string                   `json:"eventSourceArn"`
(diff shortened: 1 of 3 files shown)


bmoffatt:
More relationships:
* Similar fix was recently made for the Lambda .NET project https://github.com/aws/aws-lambda-dotnet/pull/1476
* Fixes: https://github.com/aws/aws-lambda-go/issues/422

bmoffatt:
So, since this field was never functional since being added in #385, my gut reaction is that I'd be OK merging this as-is and calling it a bug fix, rather than a semver breaking change.

However, https://github.com/aws/aws-lambda-dotnet/pull/1476#issuecomment-1498213789, makes me think it'd be possible to keep strict back-compatability by adding a custom Marshal/Unmarshal JSON to this struct, so that the field type may stay as `[]byte`. This might also be more convienient to work with for consumers, as `[]int8` isn't really a common type to be playing around with in application code. Custom Marshaling isn't abnormal for events in this library (ex: recently merged a custom marshaling in #502!)

@milovacb WDYT? Does it look feasible to do a bit of extra work to allow Go consumers to consume the kafka header value as `[]byte`, rather than having to convert it themselves? Or is there some Kafka weirdness (I'm far from an expert) where this needs to be handled as signed bytes within application code too?

milovacb:
I researched this a little bit in more detail. Official Kafka Go library does use `byte` [in its implementation](https://github.com/confluentinc/confluent-kafka-go/blob/4abbbd4c2ed8c5b5b1a2b4c23d08f8c8bddf3c2f/kafka/header.go#L44). However, this is still going to be used inside of a Lambda function invoked by ESM (Event Source Mapping) and negative values may arrive.

It comes down to customer experience - should this library match the experience of Kafka Go library or should it match Lambda experience? My experience so far was that Lambda experience was more important for customers, but I don't think I have enough data for this to make that decision and this being an open source library I guess we arrive to conclusions in a different way.

I think users can make an appropriate conversion by themselves if needed, but if it is a blocker I can make conversion to `byte` by adding 256 to all negative values (thus all negative values with 1 as the highest bit will be translated to 128-255 range)?

susan-hlu (dismissed):

bmoffatt:
I wouldn't expect a consumer to ever interact with the "negative numbers" directly. The producer, regardless of how the particular producer chooses to pretty-print it's bytes, defiantly chose to write bytes. So the consumer should get bytes too!

I think something like this would result in a more ergonomic event type

```
type KafkaFooBar struct {
   //...
   Headers map[string]JSONNumberBytes
   //....
}

// JSONNumberBytes is bytes that are json encoded/decoded as numbers in the range [-128, 127]
type JSONNumberBytes []byte

func (b JSONNumberBytes) MarshalJSON() ([]byte, error) {
	n := make([]int8, len(b))
	for i, v := range b {
		n[i] = int8(v)
	}
	return json.Marshal(n)
}

func (b *JSONNumberBytes) UnmarshalJSON(p []byte) error {
	var n []int8
	if err := json.Unmarshal(p, &n); err != nil {
		return err
	}
	*b = make(JSONNumberBytes, len(n))
	for i, v := range n {
		(*b)[i] = byte(v)
	}
	return nil
}
```

https://go.dev/play/p/7_10l_AIsxF

milovacb:
OK, keeping bytes then. Test updated.

bmoffatt (approved):
Thanks!

PR #513, opened 2023-07-19
feat: provide the options to set arbitrary json encoder and decoder options

MatthewDolan:
*Issue #:* https://github.com/aws/aws-lambda-go/issues/514

*Description of changes:*

Before this change, the lambda API supported setting some of the json encoder options (`SetEscapeHTML(...)` & `SetIndent(...)`). I needed the ability to set the `UseNumber()` option on the decoder and I considered setting it in the same way (by creating a one-off option), but instead I ended up creating two new options that allow setting arbitrary fields on the encoder and decoder.

I think this is a better option because it's future-proof (if new options are added to the json library, this library won't also need to be changed to make this option available) and more backward compatible.

### DIFF
--- a/lambda/handler.go
+++ b/lambda/handler.go
@@ -25,6 +25,7 @@ type handlerOptions struct {
 	baseContext        context.Context
 	jsonEncoderOptions []func(encoder *json.Encoder)
 	jsonDecoderOptions []func(decoder *json.Decoder)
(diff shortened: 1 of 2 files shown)


MatthewDolan:
@bmoffatt, I would be curious to hear your feedback on this change. Thank you.

PR #519, opened 2023-09-22
Update README.md

luizarnoldch:
I found a bug in my linux compilation, When i compile the code with main.go in the same directory, i get an error of compiling. So i want to help all who wants to aumatizate and only get the binary code.

*Issue #, if available:*

*Description of changes:*


bmoffatt:
This project's README is fairly minimal, and isn't the right place to add troubleshooting steps. If you've come across an error in the one of the walkthroughs on docs.aws.amazon.com, you can provide feedback via the "Provide feedback" link at the bottom of most pages.

PR #524, opened 2023-10-16
feat: apply `$AWS_LAMBDA_EXEC_WRAPPER` if set

### LINKED ISSUE #523: Support for `AWS_LAMBDA_EXEC_WRAPPER`

Datadog's ASM support for Lambda is implemented using a proxy around the `AWS_LAMBDA_RUNTIME_API` so we can intercept requests & responses and pass them to the WAF (for monitoring and possibly blocking).

This flow requires replacing the value of `AWS_LAMBDA_RUNTIME_API` with an alternate endpoint, which is normally achieved by setting the `AWS_LAMBDA_EXEC_WRAPPER` environment variable on the Lambda function to `/opt/datadog_wrapper` which then sets `AWS_LAMBDA_RUNTIME_API` to the `host:port` of the reverse proxy started by the Datadog extension.

Unfortunately, the provided runtimes (`provided`, `provided.al2`, and coincidentally `go1.x`) do not actually honor this environment variable and require special casing.

It's unclear why custom/provided runtimes fail to honor this setting, and we would like to understand if it would be possible for the provided runtimes to actually do this.

Failing that, we think it would be feasible (although perhaps not desirable, notably because of the added cold start latency) for `github.com/aws/aws-lambda-go` to come with an `init` function that detects the `AWS_LAMBDA_EXEC_WRAPPER` environment variable, and when present, uses `syscall.exec` to enact the other runtime's flow.

We are considering a focalized solution in https://github.com/DataDog/datadog-lambda-go/pull/143, but feel like this would be best addressed higher up in the dependency/responsibility chain.

**Additional context**

See: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-modify.html#runtime-wrapper.
See: https://github.com/DataDog/datadog-lambda-extension/blob/main/scripts/datadog_wrapper.

RomainMuller:
Managed AWS Lambda runtimes wrap the function's runtime entry using a script specified by the `AWS_LAMBDA_EXEC_WRAPPER` environment variable, however this is not performed by provided runtimes (`provided`, `provided.al2`, `go1.x`).

This adds a new `WithEnableExecWrapper()` option that re-startes the current process after wrapping it in the wrapper to emulate the behavior of other lambda runtimes, although this will cause initialization of some other go packages to be run twice, which will have an impact on cold start times.

See also: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-modify.html#runtime-wrapper

Fixes #523

---

### DIFF
--- a/lambda/wrapper.go
+++ b/lambda/wrapper.go
@@ -13,28 +13,27 @@ package lambda
 import (
 	"log"
 	"os"
-	"syscall"
 )
 
 const awsLambdaExecWrapper = "AWS_LAMBDA_EXEC_WRAPPER"
 
-func init() {
-	// Honor the AWS_LAMBDA_EXEC_WRAPPER configuration at startup, trying to emulate
-	// the behavior of managed runtimes, as this configuration is otherwise not applied
-	// by provided runtimes (or go1.x).
-	execAWSLambdaExecWrapper(os.Getenv, syscall.Exec)
-}
-
+// execAWSLambdaExecWrapper applies the AWS_LAMBDA_EXEC_WRAPPER environment variable.
 // If AWS_LAMBDA_EXEC_WRAPPER is defined, replace the current process by spawning
 // it with the current process' arguments (including the program name). If the call
 // to syscall.Exec fails, this aborts the process with a fatal error.
 func execAWSLambdaExecWrapper(
 	getenv func(key string) string,
 	sysExec func(argv0 string, argv []string, envv []string) error,
+	callbacks []func(),
 ) {
 	wrapper := getenv(awsLambdaExecWrapper)
 	if wrapper == "" {
 		return
+	}
+
+	// Execute the provided callbacks before re-starting the process...
+	for _, callback := range callbacks {
+		callback()
 	}
 
 	// The AWS_LAMBDA_EXEC_WRAPPER variable is blanked before replacing the process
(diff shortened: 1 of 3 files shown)


lambda/wrapper.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"log"
+	"os"
+	"syscall"
+)
+
+const awsLambdaExecWrapper = "AWS_LAMBDA_EXEC_WRAPPER"
+
+func init() {
```
bmoffatt, line 21: I think it'd be problematic to introduce this as an `init()` function. As functions with possibly vestigial definitions of AWS_LAMBDA_EXEC_WRAPPER would see a behavior change.

So if this was to be added, I think it'd best to do it as a `lambda.Option`. eg: `lambda.StartWithOptions(..., lambda.WithEnableExecWrapper())`

bmoffatt:
I'm not sure about this feature. The feature could cause issues if a current or future version or `provided` started including built-in support for AWS_LAMBDA_EXEC_WRAPPER. Will forward the proposal on to the product team.

lambda/wrapper.go
RomainMuller: > So if this was to be added, I think it'd best to do it as a lambda.Option. eg: lambda.StartWithOptions(..., lambda.WithEnableExecWrapper())

I like this idea a lot!

yoelb:
Any update regarding feature request?

Issue #528, opened 2023-11-01
events.CodeCommitReference has missing Deleted field


praveenk007:
**Is your feature request related to a problem? Please describe.**
The field `Deleted` is missing within `events.CodeCommitReference` struct. 
Due to the lack of this field, our lambda function is unable to identify if the event was of a ref deletion or not.

**Describe the solution you'd like**
Add the field in below struct
```
type CodeCommitReference struct {
	Commit  string `json:"commit"`
	Ref     string `json:"ref"`
	Created bool   `json:"created,omitempty"`
        Deleted bool   `json:"deleted,omitempty"`
}
```

and modify `String()` function as below
```
func (r CodeCommitReference) String() string {
	return fmt.Sprintf(
      "{commit: %v, ref: %v, created: %v, deleted: %v}", r.Commit, r.Ref, r.Created, r.Deleted)
}
```

**Describe alternatives you've considered**
We have been provided a patch by AWS team for now with the same above changes which we are using in our local env and CI/CD.

**Additional context**
Here is the event received by lambda function upon tag deletion.
```
{
    "Records": [
        {
            "awsRegion": "us-east-2",
            "codecommit": {
                "references": [
                    {
                        "commit": "f76677c0a0bc68567bb1d629dcbc6f753c1ae993",
                        "deleted": true,
                        "ref": "refs/tags/1.0.0"
                    }
                ]
            },
            "customData": null,
            "eventId": "f46608e0-077d-455d-978b-61bfbb34c533",
            "eventName": "ReferenceChanges",
            "eventPartNumber": 1,
            "eventSource": "aws:codecommit",
            "eventSourceARN": "***redacted",
            "eventTime": "2023-05-08T05:24:10.997+0000",
            "eventTotalParts": 1,
            "eventTriggerConfigId": "b5d6b7e6-0275-4250-b296-db33d2b880ce",
            "eventTriggerName": "MyNotification",
            "eventVersion": "1.0",
            "userIdentityARN": "***redacted"
        }
    ]
}
```

The deleted field from above json doesn't get mapped due to missing field in the struct.

bmoffatt:
Do you have a reference handy to the AWS documentation for this field?

praveenk007:
@bmoffatt There isn't any as the field itself is missing. However, here is the event received by lambda function upon tag deletion.
```
{
    "Records": [
        {
            "awsRegion": "us-east-2",
            "codecommit": {
                "references": [
                    {
[...]

PR #531, opened 2023-11-08
Added Deleted field to support codecommit ref delete events

praveenk007:
*Issue #, if available:*
#528 

*Description of changes:*
Added `Deleted` field so that the consumer is able to identify reference delete events.


leepa:
You've committed your .idea folder. You'll want to remove that.

Issue #525, opened 2023-10-24
S3ObjectLambdaEvent is missed in codebase


ikovalyov:
**Is your feature request related to a problem? Please describe.**
S3ObjectLambdaEvent is used for S3 Object Lambdas to communicate with S3 but this event type is missed. I can see this event exists in typescript, java, python versions of almbda runtime,

PR #536, opened 2023-12-17
Add S3ObjectLambdaEvent

kdnakt:
*Issue #, if available:*

#525

*Description of changes:*

I used JSON files for testing, similar to the repository below.
https://github.com/awslabs/aws-lambda-rust-runtime/tree/main/lambda-events/src/fixtures

### DIFF
--- a/events/README_S3_Object_Lambda.md
+++ b/events/README_S3_Object_Lambda.md
@@ -1,0 +1,78 @@
+# Sample Function
+
+The following is a sample class and Lambda function that receives Amazon S3 Object Lambda event record data as an input and returns an object metadata output.
+
+```go
+
+// main.go
+package main
+
+import (
+	"context"
+	"crypto/md5"
+	"encoding/hex"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"github.com/aws/aws-lambda-go/lambda"
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+)
+
+func handler(ctx context.Context, event events.S3ObjectLambdaEvent) (*s3.WriteGetObjectResponseOutput, error) {
+	url := event.GetObjectContext.InputS3Url
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	transformedObject := TransformedObject{
+		Metadata: Metadata{
+			Length: len(bodyBytes),
+			Md5:    toMd5(bodyBytes),
+		},
+	}
+	jsonData, err := json.Marshal(transformedObject)
+	if err != nil {
+		return nil, err
+	}
+	cfg, err := config.LoadDefaultConfig(context.TODO())
+	if err != nil {
+		return nil, err
+	}
+	svc := s3.NewFromConfig(cfg)
+	input := &s3.WriteGetObjectResponseInput{
+		RequestRoute: &event.GetObjectContext.OutputRoute,
+		RequestToken: &event.GetObjectContext.OutputToken,
+		Body:         strings.NewReader(string(jsonData)),
+	}
+	return svc.WriteGetObjectResponse(ctx, input)
+}
+
+func toMd5(data []byte) string {
+	hasher := md5.New()
+	hasher.Write(data)
+	hash := hasher.Sum(nil)
+
+	return hex.EncodeToString(hash)
+}
+
+type TransformedObject struct {
+	Metadata Metadata `json:"metadata"`
+}
+
+type Metadata struct {
+	Length int    `json:"length"`
+	Md5    string `json:"md5"`
+}
+
+func main() {
+	lambda.Start(handler)
+}
+
+```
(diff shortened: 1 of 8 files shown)


events/s3_object_lambda.go
```diff
@@ -0,0 +1,64 @@
...
+
+type ListObjectsV2Context struct {
+	InputS3Url string `json:"inputS3Url"`
+}
+
+type HeadObjectContext struct {
+	InputS3Url string `json:"inputS3Url"`
+}
+
+type Configuration struct {
```
bmoffatt, line 33: All the new structs in the PR should be prefixed with something like `S3ObjectLambda`, to avoid polluting the shared event namespace with very generic names like `Configuration`.

bmoffatt (changes requested):

events/README_S3_Object_Lambda.md
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, err
+	}
+	svc := s3.NewFromConfig(cfg)
+	input := &s3.WriteGetObjectResponseInput{
+		RequestRoute: &event.GetObjectContext.OutputRoute,
+		RequestToken: &event.GetObjectContext.OutputToken,
+		Body:         strings.NewReader(string(jsonData)),
+	}
+	return svc.WriteGetObjectResponse(ctx, input)
```
bmoffatt, line 54: At a glance this doesn't look right as an example return value.

Looking at the docs https://docs.aws.amazon.com/AmazonS3/latest/userguide/olap-writing-lambda.html#olap-headobject

I see python functions returning things like `{statusCode: 200, <otherfields>}` and at least on java function defined with a `void` return type.

events/s3_object_lambda.go
kdnakt: Thanks for the suggestion, I added the prefix.

events/README_S3_Object_Lambda.md
kdnakt: Changed this function just to return error.

events/s3_object_lambda.go
```diff
@@ -0,0 +1,64 @@
...
+	SessionIssuer *SessionIssuer    `json:"sessionIssuer,omitempty"`
+}
+
+type SessionIssuer struct {
+	Type        string `json:"type"`
+	PrincipalID string `json:"principalId"`
+	ARN         string `json:"arn"`
+	AccountID   string `json:"accountId"`
+	UserName    string `json:"userName"`
+}
```
bmoffatt, line 64: All of these structs should be prefixed with `S3ObjectLambda`

kdnakt: I fixed them!

events/s3_object_lambda.go
```diff
@@ -0,0 +1,64 @@
+package events
+
+type S3ObjectLambdaEvent struct {
+	XAmzRequestID        string                              `json:"xAmzRequestId"`
+	GetObjectContext     *S3ObjectLambdaGetObjectContext     `json:"getObjectContext,omitempty"`
+	ListObjectsContext   *S3ObjectLambdaListObjectsContext   `json:"listObjectsContext,omitempty"`
+	ListObjectsV2Context *S3ObjectLambdaListObjectsV2Context `json:"listObjectsV2Context,omitempty"`
+	HeadObjectContext    *S3ObjectLambdaHeadObjectContext    `json:"headObjectContext,omitempty"`
```
bmoffatt, line 8: Looking for another reference for these fields. The documentation I found so far only shows `getObjectContext`

https://docs.aws.amazon.com/AmazonS3/latest/userguide/olap-event-context.html

events/s3_object_lambda.go
```diff
@@ -0,0 +1,64 @@
...
+	Configuration        S3ObjectLambdaConfiguration         `json:"configuration"`
+	UserRequest          S3ObjectLambdaUserRequest           `json:"userRequest"`
+	UserIdentity         S3ObjectLambdaUserIdentity          `json:"userIdentity"`
+	ProtocolVersion      string                              `json:"protocolVersion"`
+}
+
+type S3ObjectLambdaGetObjectContext struct {
+	InputS3Url  string `json:"inputS3Url"`
```
bmoffatt, line 16: Linter should have flagged this 🧐 Should be `InputS3URL` instead of `InputS3Url`

events/s3_object_lambda.go
bmoffatt: PR for the rust event https://github.com/calavera/aws-lambda-events/pull/127 didn't contain additional docs for List/Head object.

Guess I gotta check this by hand, or ask the service team for a pointer.

events/s3_object_lambda.go
kdnakt: Thanks, I fixed them.

events/s3_object_lambda.go
kdnakt: https://docs.aws.amazon.com/AmazonS3/latest/userguide/olap-writing-lambda.html#olap-getobject-response

I followed these examples in the documentation.

bmoffatt: My bad! I forgot that I referenced that exact documentation when reviewing one of your previous commits! https://github.com/aws/aws-lambda-go/pull/536#discussion_r1434608488

bmoffatt (approved):
LGTM, thanks!

events/s3_object_lambda.go
kdnakt: Thanks for letting me know!
I'll create another pull request for that.

PR #538, opened 2023-12-19
Added support for Cognito pre token generation with access token customization

AMZ-brandon:
*Description of changes:*
Added CognitoEventUserPoolsPreTokenGenV2 event context. Version 2 of the pre token generation trigger introduces the ability to customize the access token. See https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-token-generation.html#cognito-user-pools-lambda-trigger-syntax-pre-token-generation

### DIFF
--- a/events/cognito.go
+++ b/events/cognito.go
@@ -137,7 +137,7 @@ type CognitoEventUserPoolsPreTokenGenRequest struct {
 type CognitoEventUserPoolsPreTokenGenV2Request struct {
 	UserAttributes     map[string]string  `json:"userAttributes"`
 	GroupConfiguration GroupConfiguration `json:"groupConfiguration"`
-	ClientMetadata     map[string]string  `json:"clientMetadata"`
+	ClientMetadata     map[string]string  `json:"clientMetadata,omitempty"`
 	Scopes             []string           `json:"scopes"`
 }
 
--- a/events/cognito.go
+++ b/events/cognito.go
@@ -2,7 +2,7 @@
 
 package events
 
-// CognitoEvent contains data from an event sent from AWS Cognito Sync
+// CognitoEvent contains data from an event sent from Amazon Cognito Sync
 type CognitoEvent struct {
 	DatasetName    string                          `json:"datasetName"`
 	DatasetRecords map[string]CognitoDatasetRecord `json:"datasetRecords"`
@@ -13,14 +13,14 @@ type CognitoEvent struct {
 	Version        int                             `json:"version"`
 }
 
-// CognitoDatasetRecord represents a record from an AWS Cognito Sync event
+// CognitoDatasetRecord represents a record from an Amazon Cognito Sync event
 type CognitoDatasetRecord struct {
 	NewValue string `json:"newValue"`
 	OldValue string `json:"oldValue"`
 	Op       string `json:"op"`
 }
 
-// CognitoEventUserPoolsPreSignup is sent by AWS Cognito User Pools when a user attempts to register
+// CognitoEventUserPoolsPreSignup is sent by Amazon Cognito User Pools when a user attempts to register
 // (sign up), allowing a Lambda to perform custom validation to accept or deny the registration request
 type CognitoEventUserPoolsPreSignup struct {
 	CognitoEventUserPoolsHeader
@@ -28,7 +28,7 @@ type CognitoEventUserPoolsPreSignup struct {
 	Response CognitoEventUserPoolsPreSignupResponse `json:"response"`
 }
 
-// CognitoEventUserPoolsPreAuthentication is sent by AWS Cognito User Pools when a user submits their information
+// CognitoEventUserPoolsPreAuthentication is sent by Amazon Cognito User Pools when a user submits their information
 // to be authenticated, allowing you to perform custom validations to accept or deny the sign in request.
 type CognitoEventUserPoolsPreAuthentication struct {
 	CognitoEventUserPoolsHeader
@@ -36,7 +36,7 @@ type CognitoEventUserPoolsPreAuthentication struct {
 	Response CognitoEventUserPoolsPreAuthenticationResponse `json:"response"`
 }
 
-// CognitoEventUserPoolsPostConfirmation is sent by AWS Cognito User Pools after a user is confirmed,
+// CognitoEventUserPoolsPostConfirmation is sent by Amazon Cognito User Pools after a user is confirmed,
 // allowing the Lambda to send custom messages or add custom logic.
 type CognitoEventUserPoolsPostConfirmation struct {
 	CognitoEventUserPoolsHeader
@@ -44,7 +44,7 @@ type CognitoEventUserPoolsPostConfirmation struct {
 	Response CognitoEventUserPoolsPostConfirmationResponse `json:"response"`
 }
 
-// CognitoEventUserPoolsPreTokenGen is sent by AWS Cognito User Pools when a user attempts to retrieve
+// CognitoEventUserPoolsPreTokenGen is sent by Amazon Cognito User Pools when a user attempts to retrieve
 // credentials, allowing a Lambda to perform insert, suppress or override claims
 type CognitoEventUserPoolsPreTokenGen struct {
 	CognitoEventUserPoolsHeader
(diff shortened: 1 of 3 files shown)


events/cognito.go
```diff
@@ -13,54 +13,62 @@ type CognitoEvent struct {
 	Version        int                             `json:"version"`
 }
 
-// CognitoDatasetRecord represents a record from an AWS Cognito Sync event
+// CognitoDatasetRecord represents a record from an Amazon Cognito Sync event
```
bmoffatt, line 16: thanks for the doc fixes too!

events/cognito.go
```diff
@@ -125,11 +133,24 @@ type CognitoEventUserPoolsPreTokenGenRequest struct {
 	ClientMetadata     map[string]string  `json:"clientMetadata"`
 }
 
-// CognitoEventUserPoolsPreTokenGenResponse containst the response portion of  a PreTokenGen event
+// CognitoEventUserPoolsPreTokenGenV2Request contains request portion of V2 PreTokenGen event
+type CognitoEventUserPoolsPreTokenGenV2Request struct {
+	UserAttributes     map[string]string  `json:"userAttributes"`
+	GroupConfiguration GroupConfiguration `json:"groupConfiguration"`
```
bmoffatt, line 139: https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-token-generation.html#cognito-user-pools-lambda-trigger-syntax-pre-token-generation example says this is an array. 

`
...
        "groupConfiguration": [{
            "groupsToOverride": ["string", "string"],
            "iamRolesToOverride": ["string", "string"],
            "preferredRole": "string"
        }],
...
`

but the prose describes it as an object, and also states that it's the same as the v1 event, which in this library is also modeled as an object
> The input object that contains the current group configuration. The object includes groupsToOverride, iamRolesToOverride, and preferredRole.

🧐 Any chance you know if the docs are wrong on this one? A sample event from elsewhere in the docs that we could paste into the test cases too would be helpful

events/cognito.go
```diff
@@ -125,11 +133,24 @@ type CognitoEventUserPoolsPreTokenGenRequest struct {
...
+}
+
+// CognitoEventUserPoolsPreTokenGenResponse contains the response portion of a PreTokenGen event
 type CognitoEventUserPoolsPreTokenGenResponse struct {
 	ClaimsOverrideDetails ClaimsOverrideDetails `json:"claimsOverrideDetails"`
 }
 
+// CognitoEventUserPoolsPreTokenGenV2Response contains the response portion of a V2 PreTokenGen event
+type CognitoEventUserPoolsPreTokenGenV2Response struct {
+	ClaimsAndScopeOverrideDetails ClaimsAndScopeOverrideDetails `json:"claimsAndScopeOverrideDetails"`
```
bmoffatt, line 151: I see this field in the docs schema exmaple. But not in the table describing the fields.

https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-token-generation.html#cognito-user-pools-lambda-trigger-syntax-pre-token-generation

Will need to follow-up with the service or documentation team to verify the correct schema.

Copypasting table as I see it right now incase there's churn

---

Name 	Description 	Minimum trigger event version
idTokenGeneration 	

The claims that you want to override, add, or suppress in your user’s ID token. This parent to ID token customization values appears only in version 2 events, but the children elements appear in version 1 events.
	2
accessTokenGeneration 	

The claims and scopes that you want to override, add, or suppress in your user’s access token. This parent to access token customization values appears only in version 2 events.
	2
claimsToAddOrOverride 	

A map of one or more claims and their values that you want to add or modify. For group-related claims, use groupOverrideDetails instead.

In version 2 events, this element appears under both accessTokenGeneration and idTokenGeneration.
	1
claimsToSuppress 	

A list of claims that you want Amazon Cognito to suppress. If your function both suppresses and replaces a claim value, then Amazon Cognito suppresses the claim.

In version 2 events, this element appears under both accessTokenGeneration and idTokenGeneration.
	1
groupOverrideDetails 	

The output object that contains the current group configuration. The object includes groupsToOverride, iamRolesToOverride, and preferredRole.

Your function replaces the groupOverrideDetails object with the object that you provide. If you provide an empty or null object in the response, then Amazon Cognito suppresses the groups. To keep the existing group configuration the same, copy the value of the groupConfiguration object of the request to the groupOverrideDetails object in the response. Then pass it back to the service.

Amazon Cognito ID and access tokens both contain the cognito:groups claim. Your groupOverrideDetails object replaces the cognito:groups claim in access tokens and ID tokens.
	1
scopesToAdd 	

A list of OAuth 2.0 scopes that you want to add to the scope claim in your user's access token. You can't add scope values that contain one or more blank-space characters.
	2
scopesToSuppress 	

A list of OAuth 2.0 scopes that you want to remove from the scope claim in your user's access token.
	2

---

events/cognito.go
bmoffatt: Got a sample event from Brandon - turns out this one is a documentation issue.

events/cognito.go
bmoffatt: Got updated sample data - table looks correct for the fields is has, and sample data verifies the presence of `claimsAndScopeOverrideDetails`

events/testdata/cognito-event-userpools-pretokengen-v2.json
```diff
@@ -0,0 +1,70 @@
...
+          "group-A",
+          "group-B",
+          "group-C"
+        ],
+        "iamRolesToOverride": [
+          "arn:aws:iam::123456789012:role/sns_callerA",
+          "arn:aws:iam::123456789012:role/sns_callerB",
+          "arn:aws:iam::123456789012:role/sns_callerC"
+        ],
+        "preferredRole": "arn:aws:iam::123456789012:role/sns_caller"
```
bmoffatt, line 66: ok so this one is missing in the structs per the test output

```
         	            	--- Expected
        	            	+++ Actual
        	            	@@ -6,4 +6,5 @@
        	            	  (string) (len=6) "region": (string) (len=9) "us-west-2",
        	            	- (string) (len=7) "request": (map[string]interface {}) (len=3) {
        	            	-  (string) (len=18) "groupConfiguration": (map[string]interface {}) (len=2) {
        	            	+ (string) (len=7) "request": (map[string]interface {}) (len=4) {
        	            	+  (string) (len=14) "clientMetadata": (interface {}) <nil>,
        	            	+  (string) (len=18) "groupConfiguration": (map[string]interface {}) (len=3) {
        	            	    (string) (len=16) "groupsToOverride": ([]interface {}) {
        	            	@@ -11,3 +12,4 @@
        	            	    (string) (len=18) "iamRolesToOverride": ([]interface {}) {
        	            	-   }
        	            	+   },
        	            	+   (string) (len=13) "preferredRole": (interface {}) <nil>
        	            	   },
```

bmoffatt: and `clientMetadata` is missing in the sample event, though it's present in the docs and the v1 event

and docs make it look pretty optional


> To pass this data to your Lambda function, use the ClientMetadata parameter in the [AdminRespondToAuthChallenge](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminRespondToAuthChallenge.html) and [RespondToAuthChallenge](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_RespondToAuthChallenge.html) API operations. Amazon Cognito doesn't include data from the ClientMetadata parameter in [AdminInitiateAuth](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminInitiateAuth.html) and [InitiateAuth](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_InitiateAuth.html) API operations in the request that it passes to the pre token generation function.

so an `omitempty` annotation should make the ser-de test happy, and not make the sample data any less correct

bmoffatt: misread the test output of the `preferredRole`, it's that the struct was re-used in a context that didn't have it populated, and serialization of the struct wrote a `null` instead of omitting the value

need to tweak the test a bit, so things that expect `prefferedRole: null` don't barf

bmoffatt: Went with adding a `null` entry to the test event instead of leaving if undefined. This lets the marshal/unmarshal test stay in the same style as the other tests.

bmoffatt (approved):

PR #546, opened 2024-01-11
lambdaurl: Populate RemoteAddr with SourceIP

lyoung-confluent:
Currently when using the `lamdbaurl` package the `RemoteAddr` field of the parsed `*http.Request` is not populated:
```go
// RemoteAddr allows HTTP servers and other software to record
// the network address that sent the request, usually for
// logging. This field is not filled in by ReadRequest and
// has no defined format. The HTTP server in this package
// sets RemoteAddr to an "IP:port" address before invoking a
// handler.
// This field is ignored by the HTTP client.
RemoteAddr string
```
This PR changes the behavior to populate the value from the underlying request context. This matches the behavior of other AWS libraries: https://github.com/awslabs/aws-lambda-go-api-proxy/blob/09c9b90f1fd1a45c05eb3a2a7a166526fcddabd8/core/requestv2.go#L168

### DIFF
--- a/lambdaurl/http_handler.go
+++ b/lambdaurl/http_handler.go
@@ -118,6 +118,7 @@ func Wrap(handler http.Handler) func(context.Context, *events.LambdaFunctionURLR
 		if err != nil {
 			return nil, err
 		}
+		httpRequest.RemoteAddr = request.RequestContext.HTTP.SourceIP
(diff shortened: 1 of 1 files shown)


bmoffatt (approved):

bmoffatt:
Nice, thanks!

PR #550, opened 2024-01-25
Add S3 Batch Operations 2.0 Schema Fields to S3BatchJob/S3BatchJobTask

lyoung-confluent:
This PR adds support for the new S3 Batch Job 2.0 Invocation Schema by adding the `userArguments` field and the `s3Bucket` key (instead of `s3BucketArn`) to the existing `S3BatchJob` and `S3BatchJobTask` structs. For more details see this blog post: https://aws.amazon.com/blogs/storage/automate-object-processing-in-amazon-s3-directory-buckets-with-s3-batch-operations-and-aws-lambda/

~I'm unsure if this should be a new struct specifically for 2.0 or just a small modification to the existing struct as implemented in this PR. Lambda Powertools for Python has implemented it as the same object/class for both schema versions but also has the benefit of Python's looser typing: https://github.com/aws-powertools/powertools-lambda-python/pull/3572~


bmoffatt:
> I'm unsure if this should be a new struct specifically for 2.0 or just a small modification to the existing struct as implemented in this PR. 

Since the new schema version removes a field, rather than being purely additive, let's do new structs with a `V2` suffix.

lyoung-confluent:
@bmoffatt I've split it into a new struct `S3BatchJobV2`, however the response struct has not changed with 2.0, should I duplicate it anyway?

bmoffatt:
> @bmoffatt I've split it into a new struct `S3BatchJobV2`, however the response struct has not changed with 2.0, should I duplicate it anyway?

I'd say leave the response struct as-is

bmoffatt (approved):

PR #574, opened 2024-11-12
feat: support of otf metadata for iceberg destinations

jbleduigou:
# What?
Add support of `otfMetadata` when using Iceberg as destionations for Firehose.

# Why?
Firehose can route incoming records in a stream to different Iceberg tables based on the content of the record.   
Also it is possible to provide a `operation` attribute which will indicate to Iceberg if the record should be treated as an `insert`, `update` or `delete` operation.  
See https://docs.aws.amazon.com/firehose/latest/dev/apache-iceberg-format-input-record-different.html#apache-iceberg-route-lambda

### DIFF
--- a/events/firehose.go
+++ b/events/firehose.go
@@ -37,7 +37,14 @@ type KinesisFirehoseResponseRecord struct {
 }
 
 type KinesisFirehoseResponseRecordMetadata struct {
-	PartitionKeys map[string]string `json:"partitionKeys"`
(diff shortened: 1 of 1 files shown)


amalakar:
I am seeing this only after I submitted the PR fixing it: https://github.com/aws/aws-lambda-go/pull/576 
The issue I created: https://github.com/aws/aws-lambda-go/issues/575

Issue #575, opened 2024-11-20
Kinesis Firehose lambda api doesn't support routing to multiple table for iceberg


amalakar:
**Is your feature request related to a problem? Please describe.**
As per: https://docs.aws.amazon.com/firehose/latest/dev/apache-iceberg-format-input-record-different.html
In order to set the target table while processing via lambda needs to set metadata as:

```json
    "metadata":{
        "otfMetadata":{
            "destinationTableName":"gameevents",
            "destinationDatabaseName":"PlayerSpawn",
            "operation":"insert"
        }
    }
```
But the response metadata struct only supports partition keys https://github.com/aws/aws-lambda-go/blob/1ba09afd7523e0d5a8a2e2fb58ddb8ec8b92ebb7/events/firehose.go#L40

```golang
type KinesisFirehoseResponseRecord struct {
.....
	Metadata KinesisFirehoseResponseRecordMetadata `json:"metadata"`
}

type KinesisFirehoseResponseRecordMetadata struct {
	PartitionKeys map[string]string `json:"partitionKeys"`
}
```
**Describe the solution you'd like**
Please support `otfMetadata` as part of metadata for lambda response object.

seanhagen:
Would really love to get the associated PR finished & merged; I need this functionality for something I'm working on!

110Sw1t:
Waiting for it here as well

PR #576, opened 2024-11-21
Support otfMetadata in Kinesis firehose response metadata for iceberg table routing

amalakar:
Issue: https://github.com/aws/aws-lambda-go/issues/575

Add otfMetadata in metadata for firehose kinesis response record to support table routing.

### DIFF
--- a/events/firehose.go
+++ b/events/firehose.go
@@ -37,7 +37,14 @@ type KinesisFirehoseResponseRecord struct {
 }
 
 type KinesisFirehoseResponseRecordMetadata struct {
-	PartitionKeys map[string]string `json:"partitionKeys"`
+	PartitionKeys map[string]string                  `json:"partitionKeys"`
+	OTFMetadata   KinesisFirehoseResponseOTFMetadata `json:"otfMetadata"`
+}
+
+type KinesisFirehoseResponseOTFMetadata struct {
+	DestinationTableName    string `json:"destinationTableName"`
+	DestinationDatabaseName string `json:"destinationDatabaseName"`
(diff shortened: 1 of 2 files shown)


events/testdata/kinesis-firehose-response.json
```diff
@@ -5,15 +5,25 @@
        "recordId": "record1",
        "result": "TRANSFORMED_STATE_OK",
        "metadata": {
-         "partitionKeys": {}
+         "partitionKeys": {},
+         "otfMetadata": {
```
seanhagen, line 9: How does Firehose handle metadata with empty strings? Will this cause an error? IE, if I want to transform but *not* route, what happens?

events/firehose.go
```diff
@@ -25,6 +25,13 @@ const (
 	KinesisFirehoseTransformedStateProcessingFailed = "ProcessingFailed"
 )
 
+// Constants used for otf operation for the record
+const (
+	KinesisFirehoseOtfOperationInsert = "insert"
+	KinesisFirehoseOtfOperationUpdate = "update"
+	KinesisFirehoseOtfOperationDelete = "delete"
```
bmoffatt, line 32: Expected `type KinesisFirehoseOTFOperation string`, like is done elsewhere for similar enum-likes in other events

events/testdata/kinesis-firehose-response.json
bmoffatt: Without knowing Firehose's behavior, it'd be prudent to add 'omitempty' to the new struct tag

PR #578, opened 2024-11-28
Add Cc to common headers

bbaayyy:
*Issue #, if available:*

*Description of changes:*
During development, I confirmed that the commonHeders of the event sent from the ses included the Cc. However, I added it because I felt uncomfortable because there was no field in that structure.

### DIFF
--- a/events/ses.go
+++ b/events/ses.go
@@ -49,6 +49,7 @@ type SimpleEmailHeader struct {
 type SimpleEmailCommonHeaders struct {
 	From       []string `json:"from"`
 	To         []string `json:"to"`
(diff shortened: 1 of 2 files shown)


bmoffatt:
couldn't find a sample that included it https://docs.aws.amazon.com/ses/latest/dg/receiving-email-notifications-examples.html

but found it mentioned elsewhere in the docs https://docs.aws.amazon.com/ses/latest/dg/receiving-email-notifications-contents.html#receiving-email-notifications-contents-mail-object-commonHeaders

bmoffatt (approved):

PR #582, opened 2025-02-03
chore(dep): update testify to v1.10.0

n2p5:
*Issue #, if available:*
NA

*Description of changes:*
This started as a small PR to update `testify` to v1.10.0, but it ended up requiring a few additional changes to make the linter and test runner happy.

- REMOVED `v1.13`, `v1.14`, `v1.15`, and `v1.16` from the test runner matrix; these are no longer supported with the current version of `testify.` This should not be a problem with future-forward releases, since these are VERY OLD versions of go.
- UPDATED go.sum using `go mod tidy` to clean up old dependencies, such as `github.com/stretchr/objx` that are no longer needed.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: 'stable'
+          go-version: "stable"
       - name: golangci-golint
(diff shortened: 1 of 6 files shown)


n2p5:
(CC @bmoffatt et al) I use `aws-lambda-go` on a project and I just wanted to give it some TLC. I'd love some feedback if I need to make any changes to the PR before it is approved.

PR #622, opened 2026-03-27
fix(lambdaurl): proxy Cookies slice to http.Request header in Wrap

### LINKED ISSUE #597: lambdaurl Wrap is not proxying `LambdaFunctionURLRequest.Cookies`
The `Wrap` function from https://github.com/aws/aws-lambda-go/blob/main/lambdaurl/http_handler.go#L105 creates an `http.Request` from an `events.LambdaFunctionURLRequest`. I don't see any references to proxying the original request's cookies similar to what [aws-lambda-go-api-proxy](https://github.com/awslabs/aws-lambda-go-api-proxy/blob/master/core/requestv2.go#L169C1-L173C1) does:
```go

	for _, cookie := range req.Cookies {
		httpRequest.Header.Add("Cookie", cookie)
	}

```

This is a deal-breaker for me because my stream handlers rely on cookies

Copy and paste that code block (I'll do this with a PR after creating this issue).

I don't think this module would like to take a dependency on `aws-lambda-go-api-proxy` so copying and pasting may be best.

**Additional context**
None.

mangomaker5:
## Summary

Fixes #597

The `Wrap` function in `lambdaurl` converts a `LambdaFunctionURLRequest` into a standard Go `http.Request`. Currently, it only proxies cookies found in `Headers["cookie"]` and completely ignores the dedicated `Cookies` field on the request struct.

This breaks use cases where stream handlers rely on `r.Cookies()` and AWS populates the `Cookies` slice instead of (or in addition to) the `cookie` header.

## Changes

Added logic to proxy `request.Cookies` entries into the `http.Request` header, with a guard to prevent duplicate cookie headers when both `Headers["cookie"]` and `Cookies` are populated by AWS:

```go
hasCookie := false
for k, v := range request.Headers {
    httpRequest.Header.Add(k, v)
    if strings.ToLower(k) == "cookie" {
        hasCookie = true
    }
}
if !hasCookie {
    for _, cookie := range request.Cookies {
        httpRequest.Header.Add("Cookie", cookie)
    }
}
```

This follows the same pattern used by [aws-lambda-go-api-proxy](https://github.com/awslabs/aws-lambda-go-api-proxy), as referenced in the original issue.

## Testing

- All existing tests pass (`go test -race ./lambdaurl/...`)
- The existing test fixture `function-url-request-with-headers-and-cookies-and-text-body.json` contains both `Headers["cookie"]` and `Cookies` — the `hasCookie` guard ensures no duplication in this case
- No changes to test files were needed; existing test coverage already validates this path

## Notes

This issue has been open since October 2025 with no maintainer response or linked PRs. Since the fix is minimal (9 lines added, 0 lines removed from logic) and all existing tests pass, I've submitted this to move the conversation forward.

### DIFF
--- a/lambdaurl/http_handler.go
+++ b/lambdaurl/http_handler.go
@@ -119,8 +119,17 @@ func Wrap(handler http.Handler) func(context.Context, *events.LambdaFunctionURLR
 			return nil, err
 		}
 		httpRequest.RemoteAddr = request.RequestContext.HTTP.SourceIP
+		hasCookie := false
 		for k, v := range request.Headers {
 			httpRequest.Header.Add(k, v)
+			if strings.ToLower(k) == "cookie" {
+				hasCookie = true
+			}
+		}
+		if !hasCookie {
+			for _, cookie := range request.Cookies {
+				httpRequest.Header.Add("Cookie", cookie)
+			}
 		}
 
 		ready := make(chan header) // Signals when it's OK to start returning the response body to Lambda


bmoffatt:
Makes sense to me. Though can you speak more to the setup you have where you've observed an event payload with `"cookies":["value1","value2"]` but without a matching `"headers":{"cookie":"value1;value2"}`.

mangomaker5:
I haven't hit this in a live AWS payload myself — I picked this up from #597 and confirmed the gap in the code.

The proof is simple. Given a payload like:

```json
{
  "headers": {"content-type": "application/json"},
  "cookies": ["session_id=abc123", "user=testuser"]
}
```

Without this fix, `r.Cookies()` returns empty. `Wrap` only reads from `Headers["cookie"]` and completely ignores the `Cookies` array.

`LambdaFunctionURLRequest` defines `Cookies []string` as a first-class field — `Wrap` should handle it. This is also how `aws-lambda-go-api-proxy` handles it.

The fallback guard just prevents duplication when AWS sends both.

nguyengg:
Thanks for working on this!

Issue #58, opened 2018-03-15
Migrate type of Keys, NewImage and OldImage in DynamoDBStreamRecord to dynamodb.AttributeValue


morcmarc:
Keys, NewImage and OldImage in [DynamoDBStreamRecord][1] currently have an internal [DynamoDBAttributeValue][2] type, however this is virtually the same type as [AttributeValue][3] in "github.com/aws/aws-sdk-go/service/dynamodb" but "less usable" since it doesn't have the convenience method, like Marshal and Unmarshal.

As @cmello [pointed out][4] in #54, aws-lambda-go doesn't depend on said package at the moment but I think we should explore the idea of either migrating the type and importing the SDK as a dependency, or copying the sub-packages over to this repo (don't like this idea), so we can have the same level of functionality.

For those wondering (@moedeveloper) what to do in the meantime, I found the following solution to be working quite well: https://gist.github.com/morcmarc/35194b39184977c2c44eb9fe063889a3

[1]: https://github.com/aws/aws-lambda-go/blob/master/events/dynamodb.go#L68
[2]: https://github.com/aws/aws-lambda-go/blob/master/events/attributevalue.go#L16
[3]: https://docs.aws.amazon.com/sdk-for-go/api/service/dynamodb/dynamodbattribute/
[4]: https://github.com/aws/aws-lambda-go/issues/54#issuecomment-373241955

canni:
Just today I was doing an upgrade from NodeJS wrapper (Apex) lambdas to native go and was hit by this issue pretty hard on around ~100 lambdas, work-around works for now, but feels wrong on many levels

morcmarc:
I'm happy to make the change and submit a pull request if the AWS team is happy with that?

morcmarc:
I implemented the change, and added aws-sdk-go as a dependency. I did a quick test and noticed an increase of ~1MB in the final binary size on average, min: 0.89MB, max: 1.3MB, median: 0.95MB. 

Disclaimer: my number are probably not representative. I only tested with 6 different lambdas and numbers were rounded to 1 decimal. 

For us, this is acceptable, but I can see this being an issue for others, considering the 50MB package limit.

bnegrao:
> I implemented the change, and added aws-sdk-go as a dependency. I did a quick test and noticed an increase of ~1MB in the final binary size on average, min: 0.89MB, max: 1.3MB, median: 0.95MB.
@morcmarc , will your modification be added to the distribution? how can I be informed ?

[...]

PR #624, opened 2026-04-25
events: add UnmarshalStreamImage and ToDynamoDBJSON helpers (#58)

MukundaKatta:
## Summary
Refs #58. Adds backwards-compatible converter helpers from `events.DynamoDBStreamRecord` to the AWS SDK's `AttributeValue` shape, without changing any existing public API.

## Approach: backwards-compatible
Did not change `DynamoDBStreamRecord` field types. The recurring concern in the issue thread was the SDK v1 vs v2 split and the binary-size cost of pulling in the SDK. This PR introduces no new module deps — helpers convert via canonical DynamoDB JSON.

## What's added
- `events/dynamodb_helpers.go`:
  - `UnmarshalStreamImage(image map[string]DynamoDBAttributeValue, out interface{}) error` — direct unmarshal-into-struct that mirrors the StackOverflow workaround referenced in the issue.
  - `(DynamoDBAttributeValue).ToDynamoDBJSON() ([]byte, error)` — render to canonical DynamoDB JSON.
  - `ToDynamoDBJSONMap(map[string]DynamoDBAttributeValue) ([]byte, error)`.
  - `(DynamoDBStreamRecord).ToDynamoDBJSON() ([]byte, error)` — full envelope.
- `events/dynamodb_helpers_test.go` — 8 tests covering scalars, nested map/list, null, set types, testdata round-trip, and envelope omitempty behavior.

## Test plan
- `go build ./...` clean.
- `go vet ./events/...` clean.
- `go test ./events/...` passes (existing + 8 new).
- No existing public API touched.

### DIFF
--- a/events/dynamodb_helpers.go
+++ b/events/dynamodb_helpers.go
@@ -1,0 +1,149 @@
+// Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+package events
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// UnmarshalStreamImage unmarshals a stream image (a map of DynamoDBAttributeValue
+// keyed by attribute name, as found in DynamoDBStreamRecord.Keys, NewImage and
+// OldImage) into the provided destination value.
+//
+// The destination is decoded using the standard encoding/json package, so
+// `json` struct tags work as expected. This sidesteps the long-standing
+// limitation that DynamoDBAttributeValue is not directly compatible with the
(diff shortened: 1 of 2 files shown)


MukundaKatta:
The `TestEnableSigterm` failure on 1.15 / 1.21 looks unrelated. It's in the `lambda/` package, this PR only adds new files in `events/`. The failure is the Docker-backed Lambda runtime timing out on init (`Reset initiated: Timeout`) which I've seen as flaky on the runners. Happy to rebase if a maintainer wants a re-run.

Issue #634, opened 2026-07-22
S3Bucket struct is missing `awsGeneratedTags` field.


jeschkies:
The `S3Bucket` struct of the S3 event is missing the new `awsGeneratedTags` field:
https://github.com/aws/aws-lambda-go/blob/87e344b1d4de6b5378079a153603e22ec1c77aa1/events/s3.go#L49-L53

It seems it had been introduced recently https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html.

PR #635, opened 2026-07-22
fix(s3event): Add missing awsGeneratedTags field.

jeschkies:
*Issue #, if available:*
#634

*Description of changes:*
This change adds the missing `awsGeneratedTags` field as described in https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html. The [Java SDK](https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/eventnotifications/s3/model/S3Bucket.html) added the same field.


obs-gh-justindaines:
Could we please get a review to add the missing `awsGeneratedTags` field on S3 event notifications?

@maxday @anzheyazzz could someone take a look when you have a chance?

Thank you!

PR #637, opened 2026-08-03
Fix serialization test bugs and enable Go 1.27 compat

djedward:
*Issue #, if available:*

*Description of changes:*
The tests fail to build for Go 1.27 due to changes in the JSON library:

It exposed an existing bug in how TestMalformedJson was being used. The method is a pass-through to `json.Unmarshal` which requires a pointer. Prior to 1.27, Go returned a SyntaxError, despite the issue being the target. I updated all the calls to properly pass a pointer and perform the check they were intending to check. I added tests for this function to make Codecov happy.

Secondly, the error message in Go 1.27 for the types has changed some. This updates the tests to be a little more flexible in their matching, allowing Go 1.27 and earlier versions to pass.

TestEnableSigterm appeared to be flaky and timing out, so used a polling mechanism, increased the timeouts and pre-pulled the image to see if that resolves the flakiness (seems to have finally passed every version).

TestConcurrencyWithRIE occasionally had failures (mainly due to taking too long "5.1s not less that 4.4"), so applied similar changes as TestEnableSigterm.

Those are optional changes and I'm happy to remove them (or feel free to edit the commit to remove them). I implemented them to help get the tests to pass.


pieterocp:
**suggestion**: Expand the test runner matrix to include 1.27?

(prevent regression + prove this works with the latest version)

djedward:
When the PR was opened, it wasn't released, but it is now. I'll get it updated.

djedward:
Ok, took a bit! Unrelated to this change, the golangci-lint action was broken because it uses the 'stable' version of Go, which is now Go 1.27, so patched that all up so it would pass.

- Added Go 1.27 to the test matrix
- Fixed linter warnings about the use of deadcode/varcheck in nolint directives. They are part of 'unused'.
- Upgraded golangci-lint to latest version to match with Go 1.27 'stable'.
- Added //nolint: govet to the use of the deprecated ioutil package which are now errors during linting (fixing would require dropping <=Go 1.15 support)
- Fixing //nolint triggered code coverage issue, so I added some tests to events/tests, since they were touched. I did modify the functions to use the testing interface, allowing them to be properly mocked/tested. The interface is compatible with t.Testing, so callers did not need to be updated.

maxday:
Thanks @djedward for fixing both of the flaky tests 💯 

About the change, I think I'm fine with dropping 1.15 <= support (EOL Aug 16, 2021), what do you think @trivenay ? No a big fan of all of those linting exceptions.

Also could you bring back the shas in github actions? Dependabot understand them well, so it's fine to keep them, it's also in our policy to fix the action versions, thanks!

djedward:
Sounds good to me, I'll start with the hash locking so the versions don't upgrade. I would definitely be supportive of dropping <= 1.15, but I'll wait for consensus on that.

trivenay:
Yes if someone wants to use the ric and they are using the deprecated version then they can use older ric versions -> This is the standard procedure we follow across all runtimes

trivenay (dismissed):

djedward:
I've removed 1.13, 1.14 and 1.15 from the test matrix. I removed all the no longer necessary //nolint directives.

maxday (approved):

trivenay (approved):

PR #638, opened 2026-08-27
feat: bump min go version to 1.26

maxday:
Raises the project's minimum supported Go version to 1.26, the oldest release that is still supported by upstream Go.

## Motivation

Per the [official Go release policy](https://go.dev/doc/devel/release), each major Go release is supported only until there are two newer major releases:

> "Each major Go release is supported until there are two newer major releases."

As of now, the two most recent major releases are **Go 1.27** (latest) and **Go 1.26**. That makes **1.26 the oldest version that is not end-of-life**, Go 1.25 and earlier no longer receive fixes, including security patches.

Pinning our minimum to 1.26 keeps us on a supported toolchain and lets us drop workarounds for EOL versions.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/aws/aws-lambda-go
```
anzheyazzz, line 1: Questions: if we are releasing 2.0.0, shouldn't we make this some like `module github.com/aws/aws-lambda-go/v2`?

ref: https://go.dev/doc/modules/version-numbers

maxday: good point! actually, maybe we don't need to bump to v2, it seems that raising the minimum version of go is not a breaking change, and usually released a minor version bumps in other packages.

maxday: ie: aws-sdk is not bumping major when they bump go version (ie: https://github.com/aws/aws-sdk-go-v2/commit/d5b9ac6e3a465972c1f735ea67ab46e8fd9a6a09)

trivenay: +1 -> Let's follow go's canonical versioning

trivenay (approved):

PR #641, opened 2026-09-17
fix: bypass proxy for Runtime and Extensions API calls

benrkia:
*Issue #, if available:*

*Description of changes:*
The Runtime and Extensions API clients built a bare http.Client, which inherits http.DefaultTransport and its ProxyFromEnvironment behavior. A customer-configured HTTP(S)_PROXY then routed calls to the API endpoint through the proxy, so a proxy in the environment could break function init even though it should never affect communication with the API.

Give both clients a transport that never proxies, cloned from the default so connection-pooling and timeout defaults are preserved.


anzheyazzz (approved):
lgtm

